PR #1, opened 2022-03-23
Initial repository setup

amarthadan:
[BEC-245](https://api3dao.atlassian.net/browse/BEC-245)


.eslintignore
```diff
@@ -0,0 +1,3 @@
+**/node_modules/**
```
Siegrift: It's pretty disapointing that these ignore files do not have an option to extend gitignore (which seems a pretty reasonable default).

For that reason I've started to like putting a comment to keep them in sync. See: https://github.com/api3dao/manager-multisig/blob/main/.gitignore#L1

Also, you should probably put all of the gitignore stuff here.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: weekly
+      time: '11:00'
+    open-pull-requests-limit: 5
+    reviewers:
+      - amarthadan
```
Siegrift, line 10: Feel free to add me as well.

.eslintignore
```diff
@@ -0,0 +1,3 @@
...
+**/dist/**
+.prettierignore
```
Siegrift: Nit: \n

.prettierrc
```diff
@@ -0,0 +1,17 @@
+{
+  "bracketSpacing": true,
+  "printWidth": 120,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "es5",
+  "useTabs": false,
+  "overrides": [
```
Siegrift, line 8: I think you also want to add solidity support https://github.com/api3dao/airnode/pull/324/commits/e7c1bf033ac3e548b2bb8293a6abce5e8134a103

Siegrift (changes requested):
Some minor changes, but otherwise LGTM.

.prettierrc
amarthadan: I'll do that once it's needed.

Siegrift (approved):
LGTM

PR #12, opened 2022-04-06
Implement configuration validation & loading

amarthadan:
[BEC-251](https://api3dao.atlassian.net/browse/BEC-251)


src/config.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+// Declare originalFs outside of mockReadFileSync to prevent infinite recursion errors in mockReadFileSync.
+const originalFs = fs.readFileSync;
+
+/**
+ * Mocks the fs library if the file path includes the specified file path substring
+ * and otherwise returns the original content.
+ */
+const mockReadFileSync = (filePathSubstr: string, mockValue: string) => {
+  return jest.spyOn(fs, 'readFileSync').mockImplementation((...args) => {
```
vponline: FWIW, I had to change this to  `mockImplementationOnce` in Airkeeper because it was causing problems with loading both Airkeeper and Airnode configs.

That's not the case here but perhaps it would make sense also here if this needs to be used in other tests later on.

README.md
```diff
@@ -14,6 +14,14 @@ yarn install
 yarn build
 ```
 
+## Configuration
+
+You need to create a configuration file `config/config.json`. Take a look at `config/config.example.json` for an example
+configuration file. You can use string interpolation (with `${VAR}` syntax) for providing secrets. Secrets are read from
+the environment variables. Wehn running locally, either just with `yarn start` or via process manager, secrets are
```
vponline:
```suggestion
the environment variables. When running locally, either just with `yarn start` or via process manager, secrets are
```
Small typo

README.md
```diff
@@ -14,6 +14,14 @@ yarn install
...
+automatically loaded from `config/secrets.env` file. Take a look at `config.secrets.example.env` for an example secrets
```
vponline:
```suggestion
automatically loaded from `config/secrets.env` file. Take a look at `config/secrets.example.env` for an example secrets
```

package.json
```diff
@@ -20,18 +20,22 @@
...
     "prettier": "prettier --check \"./**/*.{js,ts,md,yml,json}\"",
     "prettier:fix": "prettier --write \"./**/*.{js,ts,md,yml,json}\" --ignore-unknown",
-    "start": "ts-node src/main.ts",
+    "start": "scripts/run-locally.sh ts-node src/main.ts",
     "test": "jest"
   },
-  "dependencies": {},
+  "dependencies": {
+    "lodash": "^4.17.21",
+    "zod": "^3.14.3"
```
vponline, line 34: Just curious why `pm2` is not included as a dependency like with Airnode? :thinking:

src/validation.ts
```diff
@@ -0,0 +1,110 @@
...
+    beaconSets: beaonSetsSchema,
+    chains: chainsSchema,
+    gateways: gatewaysSchema,
+    templates: templatesSchema,
+    triggers: triggersSchema,
+  })
+  .strict();
+
+type SchemaType<Schema extends ZodFirstPartySchemaTypes> = z.infer<Schema>;
+export type Config = SchemaType<typeof configSchema>;
```
Siegrift: Why not just use  it directly: `z.infer<typeof xyz>`

src/validation.ts
```diff
@@ -0,0 +1,110 @@
...
+  .object({
+    airnode: evmAddressSchema,
+    templateId: evmTemplateIdSchema,
+    fetchInterval: z.number().int().positive(),
+  })
+  .strict();
+
+export const beaconsSchema = z.record(evmBeaconIdSchema, beaconSchema);
+
+// Will be refined once we start supporting beacon sets
```
Siegrift: I dislike comments such these as they will inevitably be forgotten on places. I think it's better to prefix it with TODO:

src/validation.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const envVariables = {
+  CP_SELF_HOSTED_MAINNET_URL: 'https://some.self.hosted.mainnet.url',
+  CP_INFURA_MAINNET_URL: 'https://some.infura.mainnet.url',
+  CP_INFURA_ROPSTEN_URL: 'https://some.influra.ropsten.url',
+  HTTP_GATEWAY_API_KEY: '18e06827-8544-4b0f-a639-33df3b5bc62f',
+};
+
+it('successfully parses example configuration', () => {
+  const config = JSON.parse(fs.readFileSync(path.resolve(__dirname, '..', 'config', 'config.example.json'), 'utf8'));
```
Siegrift: Nit: 

```suggestion
  const config = JSON.parse(fs.readFileSync(path.join(__dirname, '..', 'config', 'config.example.json'), 'utf8'));
```

we usually use join, but resolve works as well.

src/config.test.ts
```diff
@@ -0,0 +1,134 @@
...
+const config = JSON.parse(fs.readFileSync(path.resolve(__dirname, '..', 'config', 'config.example.json'), 'utf8'));
+const envVariables = {
+  CP_SELF_HOSTED_MAINNET_URL: 'https://some.self.hosted.mainnet.url',
+  CP_INFURA_MAINNET_URL: 'https://some.infura.mainnet.url',
+  CP_INFURA_ROPSTEN_URL: 'https://some.influra.ropsten.url',
+  HTTP_GATEWAY_API_KEY: '18e06827-8544-4b0f-a639-33df3b5bc62f',
+};
+
+describe('interpolateSecrets', () => {
+  it('interplocates variables', () => {
```
Siegrift:
```suggestion
  it('interplolates variables', () => {
```

Siegrift (approved):
👍

scripts/run-locally.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# Enable automatic export of variables
+set -a
+# Source variables from secrets.env
+. $(dirname $(readlink -f "$0"))/../config/secrets.env
```
acenolaza, line 6: Is this loading secrets as env vars? If so we already have a script that might come handy https://github.com/api3dao/airkeeper/blob/main/export-aws-env.sh#L0-L1

config/config.example.json
```diff
@@ -0,0 +1,94 @@
...
+  "beacons": {
+    "0x2ba0526238b0f2671b7981fd7a263730619c8e849a528088fd4a92350a8c2f2c": {
+      "airnode": "0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace",
+      "templateId": "0xea30f92923ece1a97af69d450a8418db31be5a26a886540a13c09c739ba8eaaa",
+      "fetchInterval": 20
+    }
+  },
+  "beaconSets": {},
+  "chains": {
+    "1": {
```
acenolaza, line 11: Would it be useful in the future to have a `name` property for each chain? 🤔

acenolaza (approved):

config/config.example.json
amarthadan: It would as I have to google what the chain ID means every time :sweat_smile: I'll keep it this way for now though since that's something that should be introduced to all configuration files.

package.json
amarthadan: I don't know why is it a dependency in Airnode. I add it as a dev dependency rather than a normal dependency because we don't really need it for the run per se. I wouldn't add it at all but I wanted the option to emulate what's happening in the container environment without a container.

vponline: Ah I see, in airnode (and airkeeper) it's used to launch the server and hardhat node background services for E2E tests.

src/validation.ts
amarthadan: Tbh I just copy & pasted it from somewhere (Airkeeper?) :sweat_smile:

scripts/run-locally.sh
amarthadan: Yes, it's doing pretty much the same thing. I just wanted a POSIX-compatible way of doing that, that would work for all shells and platforms (not sure I succeded though :smile:).

PR #13, opened 2022-04-07
Implement API request fetching

Siegrift:
https://api3dao.atlassian.net/browse/BEC-247

There are a few things included in this PR:
- I found out that the jest configuration does not reset mocks which caused me a lot of pain... I've enabled the "good defaults" now
- I've separated the api fetching into three functions (`initiateFetchingBeaconData` - starts fetching data for all beacons, `fetchBeaconDataInLoop` - fetches single beacon data in a loop, `fetchBeaconData` - makes the actual api call and updates state)
- I wanted to separate the functions into own modules, but for that I needed shared state - so I've created a tiny module, which acts as an airseeker state (I've didn't include config there though)


jest.config.js
```diff
@@ -103,7 +103,7 @@ module.exports = {
   // reporters: undefined,
 
   // Automatically reset mock state before every test
-  // resetMocks: false,
+  resetMocks: true,
```
vponline, line 106: > I found out that the jest configuration does not reset mocks which caused me a lot of pain... I've enabled the "good defaults" now

Just wanted to note that this setting is not enabled in Airnode or Airkeeper and caused some confusion for me as well :sweat_smile: 

If this is enabled it means mocks/spys cannot be defined in the `describe`-block (they will return undefined) and instead need to be in a `beforeEach` or directly in the test `it`-block, more info [here](https://github.com/facebook/jest/issues/9131#issuecomment-668790615)

No need to change, just more of an FYI if you are puzzled by mocks returning undefined :smile:

package.json
```diff
@@ -27,9 +27,11 @@
     "prettier": "prettier --check \"./**/*.{js,ts,md,yml,json}\"",
     "prettier:fix": "prettier --write \"./**/*.{js,ts,md,yml,json}\" --ignore-unknown",
     "start": "scripts/run-locally.sh ts-node src/main.ts",
-    "test": "jest"
+    "test": "jest",
+    "test:debug": "jest \"fetch-beacon-data.test.ts\""
```
amarthadan: Same as in the previous PR.

src/fetch-beacon-data.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  },
+  triggers: {
+    beaconUpdates: {
+      '1': {
+        '0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC': {
+          beacons: [
+            {
+              beaconId: '0x2ba0526238b0f2671b7981fd7a263730619c8e849a528088fd4a92350a8c2f2c',
+              deviationThreshold: 0.1,
+              heartbeatInterval: 86400,
```
amarthadan:
```suggestion
              heartbeatInterval: 86_400,
```

src/fetch-beacon-data.test.ts
```diff
@@ -0,0 +1,214 @@
...
+          updateInterval: 30,
+        },
+      },
+      '3': {
+        '0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC': {
+          beacons: [
+            {
+              beaconId: '0x2ba0526238b0f2671b7981fd7a263730619c8e849a528088fd4a92350a8c2f2c',
+              deviationThreshold: 0.2,
+              heartbeatInterval: 86400,
```
amarthadan:
```suggestion
              heartbeatInterval: 86_400,
```

src/fetch-beacon-data.ts
```diff
@@ -0,0 +1,58 @@
...
+    attemptTimeoutMs: 5_000,
+    retries: infinityRetries,
+    delay: { type: 'random', minDelayMs: 0, maxDelayMs: 2_500 },
+    totalTimeoutMs: fetchInterval * 1_000,
+  });
+
+  if (!goRes.success) {
+    console.log(`Unable to call signed data gateway. Reason: "${goRes.error}"`);
+  } else {
+    updateState((state) => ({ ...state, beaconValues: { ...state.beaconValues, [beaconId]: goRes.data } }));
```
amarthadan, line 65: This `goRes` is of type `GoResult<GoResult<any, Error>, Error>`. We shouldn't wrap those go results inside one another because the `goRes.data` at the end is actually of type `GoResult` (an object) and not the resulting data which is what we want.

src/fetch-beacon-data.ts
```diff
@@ -0,0 +1,58 @@
+import { performance } from 'perf_hooks';
```
amarthadan: What is the advantage of using `performance` in comparison with normal `Date`?

src/fetch-beacon-data.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const fetchBeaconDataInLoop = async (config: Config, beaconId: string) => {
+  while (!getState().stopSignalReceived) {
+    const startTimestamp = performance.now();
+    const { fetchInterval } = config.beacons[beaconId];
+
+    await fetchBeaconData(config, beaconId);
+
+    const duration = performance.now() - startTimestamp;
+    const waitTime = Math.max(0, fetchInterval * 1_000 - duration);
```
amarthadan, line 42: I understand what's happening here but can you maybe add a comment describing why you calculate the wait time the way you do?

src/main.ts
```diff
@@ -1,44 +1,29 @@
-import * as path from 'path';
+import { join } from 'path';
```
amarthadan: I don't like this much because the name like `join` is so generic that having it prefixed like `path.join` actually helps when you're reading the code. We don't need to care about the side in this case (we're not building it for web) so I would rather stick with the prefixed version. Actually even for the `lodash`.

amarthadan (changes requested):

src/fetch-beacon-data.ts
Siegrift: Not much, it's more precise but slower. Thinking about this once more I think we are more interested in performance here rather then accuracy.

I'll revert.

src/fetch-beacon-data.ts
Siegrift: Yeah, that will be changed when the makeSignedDataGatewayRequest fn will be implemented. I initially assumed it will return go response, but I've changed that to `SignedData | null`.

I'm returning `any` type in this PR.

src/main.ts
Siegrift: Reverted.

amarthadan (approved):

PR #14, opened 2022-04-07
BEC258: Terraform ECS

acenolaza:
I'm creating this as a WIP because I wanted to get feedback as soon as possible and also because the Airseeker docket image [is not yet in docker hub](https://api3dao.atlassian.net/browse/BEC-300).

Some notes:
1. Created an ecs cluster, task definition and service to tie them together.
2. Since I'm using FARGATE I had to add a vpc resource but since we are not exposing a frontend I didn't think it was necessary to deal with internet gateways, load balancers and security groups. The simplest way I found to add a vpc is to just define a default vpc resource with default subnets.
3. Since I wanted to add logs to the task (awslogs set in the Task Definition) I had to also attach the `AmazonECSTaskExecutionRolePolicy`.
4. I've hardcoded the airnode-client docker image url just for testing until airseeker image is ready. Once that happens I guess I'll replace it with a variable.
5. I wasn't sure if we needed to deploy to different stages so I decided to keep it simple and have all terraform files in the same directory. Also I have not created any modules until I see the need for them.


terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
+# This file is maintained automatically by "terraform init".
```
amarthadan, line 1: This should not be in the repo.

terraform/network.tf
```diff
@@ -0,0 +1,15 @@
...
+  availability_zone = "${var.aws_region}a"
+}
+
+resource "aws_default_subnet" "default_subnet_b" {
+  availability_zone = "${var.aws_region}b"
+}
+
+resource "aws_default_subnet" "default_subnet_c" {
+  availability_zone = "${var.aws_region}c"
+}
```
amarthadan: I would definitely not use the default AWS resources. As from the [documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_vpc), they are not created by Terraform (if it already exists) but it's just adopted. It's kind of a shared resource and its configuration may unexpectedly change (from within another project for example). I know that the AWS networking is a bit tricky but it's not that complicated and there's a lot of tutorials on how to set the basic one (which is what we need).

terraform/providers.tf
```diff
@@ -0,0 +1,15 @@
+terraform {
```
amarthadan: I think we should use an AWS backend (S3 bucket) for storing the state file. Even without the locking to avoid the need for DynamoDB.

terraform/variables.tf
```diff
@@ -0,0 +1,16 @@
...
+variable "aws_region" {
+  description = "AWS region for deployment"
+  type        = string
+  default     = "us-east-1"
+}
+
+variable "app_environment" {
+  type        = string
+  description = "Application Environment"
+  default     = "production"
```
amarthadan: The default should not be a production, you don't want someone to deploy the production version just be accidentally running the recipes :sweat_smile:

terraform/variables.tf
```diff
@@ -0,0 +1,16 @@
+locals {
+  app_name    = "Airseeker"
```
amarthadan: Some resources require their name to be in lower case and since you're using this as part of the name it should be lower case too.

terraform/main.tf
```diff
@@ -0,0 +1,62 @@
...
+  family = "${local.app_name}-ecs-task"
+
+  container_definitions = <<DEFINITION
+  [
+    {
+      "name": "${local.app_name}-${var.app_environment}-container",
+      "image": "docker.io/api3/airkeeper-dev:a11632d57c2c22972e8863e32cc49a004e0a622a",
+      "essential": true,
+      "cpu": 256,
+      "memory": 512,
```
amarthadan: Things like these, log retention period, etc should be configurable via variables.

terraform/variables.tf
```diff
@@ -0,0 +1,16 @@
+locals {
```
amarthadan, line 1: You can create something like `resource_prefix = "${app_name}-${environment}"` and use it as a prefix for all the resources (even tags). That way, we would be able to deploy to two different environments with the same recipes (hopefully :smile:).

terraform/network.tf
amarthadan: If I understand it correctly, we need the setup described on [this documentation page](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/AWS_Fargate.html#fargate-tasks-services).
`For a Fargate task in a private subnet to pull container images, you need a NAT gateway in the subnet to route requests to the internet.`
we don't need the task to be public (there's no need to access it from outside) and therefore we also don't need any load balancers.

terraform/roles.tf
```diff
@@ -0,0 +1,26 @@
+data "aws_iam_policy_document" "ecs_task_exec_role" {
+  statement {
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["ecs-tasks.amazonaws.com", "ec2.amazonaws.com", "ecs.amazonaws.com"]
```
amarthadan: Isn't `ecs-tasks.amazonaws.com` enought?

terraform/roles.tf
```diff
@@ -0,0 +1,26 @@
...
+  tags = {
+    Name        = "${local.app_name}-ecs-taskrole"
+    Environment = var.app_environment
+    Terraform   = "true"
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "ecs_task_exec_role" {
+  role       = aws_iam_role.ecs_task_exec_role.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
```
amarthadan, line 25: I would prefer using a custom policy that allows only exactly what we need and that is writing logs. Take a look at [this Airnode one](https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/terraform/airnode/aws/modules/function/data.tf#L15), that should be enough.

amarthadan (changes requested):

terraform/variables.tf
aquarat: Like I did recently with something :shushing_face:

terraform/.terraform.lock.hcl
acenolaza: I'll remove this file from source control. I thought it was OK to add it because chainapi do have them in the repo: https://github.com/chainapi/infrastructure/blob/26f5e901baea79a4a52b8628e96230ea4950a17d/prod/.terraform.lock.hcl

Terraform docs seem to suggest to have this file in version control repo tho:  https://www.terraform.io/language/files/dependency-lock#lock-file-location

terraform/variables.tf
acenolaza: I like to live on the edge 😅

terraform/main.tf
acenolaza: [ChainAPI](https://github.com/chainapi/infrastructure/blob/4dbe6de735b19805f9fc1b54325cc09f9ce31ed0/modules/ecs/main.tf) is using 180 but [airnode-deployer](https://github.com/api3dao/airnode/blob/master/packages/airnode-deployer/terraform/airnode/aws/modules/function/main.tf#L30) uses 90 days. Which value do you think is better?

terraform/.terraform.lock.hcl
amarthadan: Interesting. I thought it shouldn't be included. Feel free to keep it then.

terraform/main.tf
amarthadan: I don't have any preference tbh. Let's say 180 just to be on the safe side.

terraform/network.tf
```diff
@@ -0,0 +1,90 @@
+resource "aws_vpc" "airseeker_aws_vpc" {
+  cidr_block           = "10.0.0.0/16"
```
amarthadan: Maybe this can be a variable too since all the other IP addresses are.

terraform/main.tf
```diff
@@ -0,0 +1,60 @@
...
+}
+
+resource "aws_ecs_service" "airkeeper_ecs_service" {
+  name            = "${local.resource_prefix}-ecs-service"
+  cluster         = aws_ecs_cluster.airseeker_ecs_cluster.arn
+  task_definition = "${aws_ecs_task_definition.airseeker_ecs_task.family}:${aws_ecs_task_definition.airseeker_ecs_task.revision}"
+  launch_type     = local.launch_type
+  desired_count   = var.ecs_application_count
+
+  network_configuration {
```
amarthadan, line 82: Don't we need a security group saying that the ECS service can access anything on the internet and nothing can access it from the internet? Have you tried making an HTTP request from inside the container?

amarthadan (changes requested):
Looks good, just a few comments :+1:

Can you run all the Terraform files through `terraform fmt` so they have the same formatting? Actually, you can take [`terraform-fmt.ts`](https://github.com/api3dao/airnode/blob/master/scripts/terraform-fmt.ts) from Airnode and add it as a part of the linting process.

You can remove the `outputs.tf` unless we want to have something outputted there.

And I think you can switch from draft to a normal PR at this point :slightly_smiling_face:

terraform/main.tf
acenolaza: This a good point and I wasn't sure so I didn't add it but what you are saying makes sense so I'll check out the docs for security groups and add it.

acenolaza:
> Looks good, just a few comments +1

I've been running terraform fmt and terraform validate manually but I think it's a good idea to have it as part of the linting process.

> Can you run all the Terraform files through `terraform fmt` so they have the same formatting? Actually, you can take [`terraform-fmt.ts`](https://github.com/api3dao/airnode/blob/master/scripts/terraform-fmt.ts) from Airnode and add it as a part of the linting process.
> 
> You can remove the `outputs.tf` unless we want to have something outputted there.

Right, I thought that I might have a use for it but I'll just remove it for now.

> And I think you can switch from draft to a normal PR at this point slightly_smiling_face

I kept it as a Draft PR because I wasn't sure if airseeker-dev was the right image to use. Also, running `docker pull api3/airseeker-dev:latest` doesn't seem to work for me. I need to use a specific task in order to be able to pull the image from docker hub. Same thing happens when I use `latest` to pull airkeeper image. Maybe @aquarat has any idea what might be happening 🤔 I did found something about having to set some config to true in docker hub in order to be able to pull using `latest`.

amarthadan:
> I kept it as a Draft PR because I wasn't sure if airseeker-dev was the right image to use. Also, running docker pull api3/airseeker-dev:latest doesn't seem to work for me. I need to use a specific task in order to be able to pull the image from docker hub. Same thing happens when I use latest to pull airkeeper image. Maybe @aquarat has any idea what might be happening thinking I did found something about having to set some config to true in docker hub in order to be able to pull using latest.

I think the problem is that there is no `@api3/airseeker-dev` tagged as `latest` :smile: Set the default to `@api3/airseeker:0.1.0` (even though it doesn't exist yet). You can pass a dev one (with commit hash) as an variable to test that it's actually working.

terraform/main.tf
```diff
@@ -0,0 +1,61 @@
...
+    Name        = "${var.app_name}-ecs-cluster"
+    Environment = var.app_environment
+  }
+
+  depends_on = [aws_cloudwatch_log_group.airseeker_aws_cloudwatch_log_group]
+}
+resource "aws_ecs_task_definition" "airseeker_ecs_task" {
+  family = "${var.app_name}-ecs-task"
+
+  container_definitions = <<DEFINITION
```
acenolaza, line 40: I've just now realized that even though the docker image was being run by ECS the logs were showing some errors while executing the code. These errors are related to the Airseeker code not being able to find the config file. These means that I think I need to make changes to this task definition in order to maybe mount a volume with the config and also add the secret file key pairs as env vars 🤔 

This is the docker command that works for me locally:

```
docker run -it --rm \
--env-file config/secrets.env \
-v "$(pwd)/config:/app/dist/config" \
api3/airseeker:0.1.0
```

would it be a good idea to add it to the README?

amarthadan: That is absolutely true, I totally missed that :man_facepalming: The idea was to mount the config directory to `/app/config`. That's on me, I haven't tried the container with the config file being loaded. Set it up for `/app/dist/config` and I'll look into fixing it later.

scripts/terraform-fmt.ts
```diff
@@ -0,0 +1,73 @@
+import * as path from 'path';
+import { exec, ExecException } from 'child_process';
+import { exit } from 'process';
+import { logger } from '@api3/airnode-utilities';
```
amarthadan: I used `airnode-utilities` because in Airnode it's already available but if we don't need it here (and I don't think we do) we shouldn't add it just for the logger.

acenolaza: I'll remove it and just use console.log since that's what we are currently doing for the rest of the app. Once we start using a logger then we'll need to come back and make the update.

acenolaza:
@amarthadan I feel like this is much better now and it's ready for final review.

amarthadan:
> @amarthadan I feel like this is much better now and it's ready for final review.

Yes, I'll review it later today :slightly_smiling_face:

README.md
```diff
@@ -57,3 +57,24 @@ yarn docker:build
 ```
 
 Resulting image is named `api3/airseeker`.
+
```
amarthadan, line 60: Can you maybe add that Docker and Terraform need to be installed on the machine before running this?

terraform/main.tf
```diff
@@ -0,0 +1,82 @@
...
+  tags = {
+    Name        = "${var.app_name}-ecs-cluster"
+    Environment = var.app_environment
+  }
+
+  depends_on = [aws_cloudwatch_log_group.airseeker_aws_cloudwatch_log_group]
+}
+
+locals {
+  env_vars = jsonencode([for tuple in regexall("(.*)=(.*)", file(var.app_secrets_file_path)) : { name = tuple[0], value = tuple[1] }])
```
amarthadan: I'm not a big fan of this but as far as I can tell from googling around it's the only Terraform-only solution. So, yeah, let's start with this and we'll see if we run into some trouble.
Just to be on the safe side, change to regex so that the variable name is not greedy as mentioned on [StackOverflow](https://stackoverflow.com/questions/59789730/how-can-i-read-environment-variables-from-a-env-file-into-a-terraform-script#comment124135133_67653301).

terraform/variables.tf
```diff
@@ -0,0 +1,88 @@
...
+variable "app_config_file_path" {
+  description = "The path to the configuration file"
+  type        = string
+  default     = "config/airseeker.json"
+}
+
+variable "app_secrets_file_path" {
+  description = "The path to the secrets file"
+  type        = string
+  default     = "../config/secrets.env"
```
amarthadan: Why are these two default paths different? The files are in the same directory, right?

terraform/variables.tf
```diff
@@ -0,0 +1,88 @@
...
+variable "app_name" {
+  description = "The name of the application"
+  type        = string
+  default     = "airseeker"
+}
+
+variable "app_environment" {
+  description = "The environment of the application"
+  type        = string
+}
```
amarthadan: I think we can have a default value of `dev`, don't you think?

amarthadan (approved):
LGTM :muscle: just some small details. I'll try deploying it tomorrow.

docker/Dockerfile.deploy
```diff
@@ -0,0 +1,11 @@
+ARG version=latest
+
+FROM api3/airseeker-dev:${version}
```
amarthadan: I think instead of having a variable for the version we can have a variable for the full `image:version` string. That way we can have a default for a non-dev image (one that is not released yet) but override it easily during the testing with a development one.

terraform/variables.tf
```diff
@@ -0,0 +1,88 @@
...
+  description = "The version of the docker image to use"
+  type        = string
+  default     = "latest"
+}
+
+variable "app_config_file_path" {
+  description = "The path to the configuration file"
+  type        = string
+  default     = "config/airseeker.json"
+}
```
amarthadan, line 35: In the current setup this file path can't be any file path but it needs to be a path within the Docker build context. Which is fine, we can't really do much about it, but there should be a warning that this is the case.

amarthadan: Also, it needs to be a relative path, not an absolute one. Which is a bit annoying...

amarthadan: Oh, I understand the difference between the path for `app_config_file_path` and `app_secrets_file_path`. That's really confusing :sweat_smile: Not sure we can do much about it though :confused:

terraform/logs.tf
```diff
@@ -0,0 +1,15 @@
...
+  tags = {
+    Name        = "${var.app_name}-ecs-log-group"
+    Environment = var.app_environment
+  }
+}
+
+resource "aws_cloudwatch_log_stream" "airseeker_aws_cloudwatch_log_stream" {
+  name           = "${local.resource_prefix}-log-stream"
+  log_group_name = aws_cloudwatch_log_group.airseeker_aws_cloudwatch_log_group.name
+}
```
amarthadan: This looks like it's not used anywhere and I don't see any logs there in AWS either.

terraform/variables.tf
acenolaza: Yes, the files are in the same directory but the contexts when accessing those files is different. For instance the context when [reading the secrets.env](https://github.com/api3dao/airseeker/blob/adda7568cb25503eff5662bfb3c40cbdd00e346e/terraform/main.tf#L29) is the `/terraform` directory (where `terraform apply` was executed). For the `airseeker.json` file the context is defined in the docker resource [here](https://github.com/api3dao/airseeker/blob/adda7568cb25503eff5662bfb3c40cbdd00e346e/terraform/main.tf#L9) so we don't need to go up a level.

terraform/variables.tf
acenolaza: I like that idea 🤝

terraform/logs.tf
acenolaza: That is weird, I was able to see the logs when clicking on the ECS service link and on the task itself too. There's a Logs tab on both pages or you can open up CloudWatch to see them

terraform/variables.tf
```diff
@@ -19,6 +19,13 @@ variable "app_name" {
...
   description = "The environment of the application"
   type        = string
+  default     = "dev"
+}
+
+variable "app_docker_image_name" {
+  description = "The name of the docker image to use"
+  type        = string
+  default     = "airseeker"
 }
```
amarthadan: I meant that there will be just one variable with a full image name, something like
```terraform
variable "app_docker_image" {
  description = "The docker image to use"
  type        = string
  default     = "api3/airseeker:latest"
}
```

amarthadan (approved):

terraform/logs.tf
amarthadan: This is how it looks like
Screenshot_20220426_122541
The first one is (I guess) automatically created by the container and it contains logs.
The second one is this one from the receipt and is empty. So the logs are there they are just not in this log stream.

acenolaza: I usually see the log stream in [here](https://us-east-1.console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/airseeker-mk01-dev-ecs-cluster/services/airseeker-mk01-dev-ecs-service/logs), [here](https://us-east-1.console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/airseeker-mk01-dev-ecs-cluster/tasks/e7610280db0e402c869b6f57cb727c4b/logs) and/or [here](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/airseeker-mk01-dev-logs/log-events/airseeker-mk01-dev$252Fairseeker-mk01-dev-container$252Fe7610280db0e402c869b6f57cb727c4b).

terraform/variables.tf
acenolaza: Done but caused that I just used 'latest' for the docker image pushed to ECR.

acenolaza:
I've rebased this branch on top of main and I started getting an issue with the use of `console.log` in `scripts/terraform-fmt.ts` which I fixed by just putting the whole `scripts` folder in the `.eslintignore` file because those scripts are only used for development.

terraform/logs.tf
amarthadan: Yes, and the first two links are just reading the logs from the third one. Which is the first one (`airseeker-mk01-dev/airseeker-mk01-dev-container/e7610280db0e402c869b6f57cb727c4b`) on my screenshot. The second one, the one you're creating in recipes (`airseeker-mk01-dev-log-stream`), is not used anywhere. There is also no reference to that resource in recipes. I think you can safely remove it

terraform/variables.tf
amarthadan: Oh, I haven't realized you're using that version someplace else. Hmm :thinking:  The images we're building can be potentially built from the same Airseeker image just with a different configuration file. But the images have unique names, right? You won't overwrite any of these images. So, tagging them as `latest` is imo fine but I would suggest adding a [`LABEL`](https://docs.docker.com/engine/reference/builder/#label) directive storing the Airseeker version and potentially a hash of the configuration file. Something like:
```dockerfile
LABEL airseeker_image=${baseImage} \
      config_sha256=${configHash}
```
Not sure that getting the config file hash will be easy though. Terraform has functions for that but the config file path is Docker context specific not Terraform specific. So, if that would be difficult to do, omit the config file hash for now and add just the Airseeker image.

acenolaza: Should I try using that hash as the image tag instead of using `latest`? My concern is that using the same Airseeker base image with a different config file will replace the existing image in ECR, right?

amarthadan: Oh, right, the name is the same. Sorry, I thought that the image is named differently every time the build is run. Never mind then. In that case, you're correct, we should have both the Airseeker image name (the base image) and the hash of the file in the tag. Not sure what characters you can put in the tag though. You won't be able to put a `/` there I presume.
But actually, we don't need to put it solely in the tag, we can name the image as well.
So, if the base image is `api3/airseeker:v0.2` and hash is `1a2b3c` we can name the newly build image `api3/airseeker:v0.2-<hash>`. Would that make sense?

acenolaza: ~I'm thinking that we would have the same issue when creating a images using the same config file but different base image.~

~Currently the ECR docker image name is `repository_url:tag`~

~I wonder if `tag` should be a hash of base image plus config file instead 🤔~

I wrote this message before I saw your last reply and I totally agree with the name structure suggested 👍🏻

acenolaza: @amarthadan Please review my last commit. I basically created an image tag with the following structure:

`api3-airseeker-dev-0.1.0-b224647`

I've also re-worked the docker context a little bit so we don't need to go up a level in the terraform vars and they both use the same path (secrets and config file)

I've also removed the unused log-stream so I feel like this might be the merge candidate 🤞🏻

amarthadan: I thought that we will put it in the image name so you don't have to "sanitize" the string but this works as well :slightly_smiling_face: 
I think this is good to go :+1:

acenolaza: I tried but I got a weird `503 X-Auth `error if I don't remove the `:` and the `/` from the tag name 🤷🏻

PR #15, opened 2022-04-08
Make signed data requests

Siegrift:
https://api3dao.atlassian.net/browse/BEC-295


src/make-request.ts
```diff
@@ -1,6 +1,56 @@
...
+export const makeSignedDataGatewayRequest = async (
+  gateway: Gateway,
+  template: Template,
+  timeoutMs: number
+): Promise<SignedData | null> => {
+  // Initiate HTTP request to each of the gateways and resolve with the data (or reject otherwise)
+  const requests = gateway.map(async (gate) => {
+    const { apiKey, url } = gate;
+    const { endpointId, parameters } = template;
+    const fullUrl = urlJoin(url, endpointId);
```
amarthadan, line 23: I really don't think we need a new dependency for doing `url + '/' + endpoint` :smile:

src/make-request.ts
```diff
@@ -1,6 +1,56 @@
-import { Gateway, Template } from './validation';
+import { go } from '@api3/promise-utils';
+import axios from 'axios';
+import urlJoin from 'url-join';
+import anyPromise from 'promise.any';
```
amarthadan, line 3: I would actually consider bumping the environment to node 16 so we can use the native [`Promise.any`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any). Or maybe at least using [`core-js`](https://github.com/zloirock/core-js) (which is mentioned in the official documentation), it feels a bit more mature. WDYT?

src/make-request.ts
```diff
@@ -1,6 +1,56 @@
...
+import { Gateway, SignedData, signedDataSchema, Template } from './validation';
 
-// TODO: https://api3dao.atlassian.net/browse/BEC-295
-export const makeSignedDataGatewayRequest = async (_gateway: Gateway, _template: Template) => {
-  return 123 as any;
+export const makeSignedDataGatewayRequest = async (
```
amarthadan: I haven't realized you want to make multiple requests here. Shouldn't we then call the function `makeSignedDataGatewayRequests` instead? Even though it returns just one response...

src/validation.ts
```diff
@@ -95,6 +95,13 @@ export const configSchema = z
   })
   .strict();
 
+export const encodedValueSchema = z.string().regex(/^0x[a-fA-F0-9]{64}$/);
+export const signatureSchema = z.string().regex(/^0x[a-fA-F0-9]+$/);
```
amarthadan: Aren't these two regexes switched? I would imagine that the signature one will be of a fixed length (64) and that the value one will be of a dynamic length. Correct me if I'm wrong.

test/fixtures.ts
```diff
@@ -0,0 +1,19 @@
+import { SignedData } from '../src/validation';
+
+// Airkeeper with signed data gateway enabled deployed on API3 account which can be used for testing.
```
amarthadan: I still don't like this and would prefer a mock server. This is nowhere documented, don't know what version of gateway it's running, don't know if I have access to check it if it's down, and so on. The mock server is like 10 lines of code at most.

src/make-request.ts
```diff
@@ -1,6 +1,56 @@
...
+  gateway: Gateway,
+  template: Template,
+  timeoutMs: number
+): Promise<SignedData | null> => {
```
amarthadan, line 18: Maybe it would make sense to return a different structure than `SignedData`. You'll change the gateway response in [AN-572](https://api3dao.atlassian.net/browse/AN-572) anyway so you can prepare for that here. If we return the new structure here right away we would have to fix just this one function. And the structure itself is kinda weird. `{timestamp: ..., value: ..., signature:...}` would make more sense. WDYT?

amarthadan (changes requested):

src/make-request.ts
Siegrift: This one is also mentioned there https://github.com/tc39/proposal-promise-any and based on the NPM downloads its fine. I don't think "maturity" context really holds in this case, since it's a single function to be implemented (not an evolving project).

> I would actually consider bumping the environment to node 16

Crossed my mind as well, but we use v14 in all tools and I'd hate switching between the versions.

src/make-request.ts
Siegrift: Yeah, we probably should. 

I also dislike the `Gateway` type, which is actually an array of `NotExactlySureHowToNameThis` (but `Gateways` refers to the record of `Gateway`).

Siegrift: Maybe: `Gateway`, `Gateways`, `GatewaysRecord` ?

src/make-request.ts
Siegrift: I'd rather defer that to a separate PR. For now I think saving the gateway response as is is the best suited for this task. For example, we might also store some other data (e.g. which gateway is the response coming from) so this interface might change couple of times.

src/make-request.ts
Siegrift: I was actually doing this without the dependency using the https://developer.mozilla.org/en-US/docs/Web/API/URL/URL but it didn't correctly handle a case when the URL didn't end with slash (which took me some time to debug).

After googling it for a few minutes, I am convinced that this is the *currently best* way to do this. without shooting myself into foot.

src/validation.ts
Siegrift: Beacons data must actually be bytes32 (string length 64) because of this https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/dapis/DapiServer.sol#L883 and for the signature length I am not sure, so I set it to `regex+` (I don't think it matters much).

I think we should require 64 length for the encoded value, but this deserves a comment. WDYT?

test/fixtures.ts
Siegrift: > This is nowhere documented, don't know what version of gateway it's running

Yeah, valid points - I think we should document it somewhere. Pinned slack post/confluence? I was thinking this gateway would be common for all api3 dev to play/test with and not only for this test.

> don't know if I have access to check it if it's down, and so on.

Yeah, that is a good point (I personally don't have access).

> The mock server is like 10 lines of code at most.

That little? :D Anyway, such test will test a very limited happy path and the mock server can get out of sync with Airnode. I'd rather treat this Airnode as an "external dependency" and go with this test even though there are these disadvantages. If this turns out to be a problem, we can remove this and implement the mock server. WDYT?

Siegrift: Btw. we are not handling Airnode gateway versions in Airseeker so the following can happen:
a) The gateway request shape changes (e.g. v2 in the URL instead of v1)
b) The gateway response changes (e.g. refactor shape of data)

I don't think we need to care about this atm, but we should in the future. WDYT?

src/make-request.ts
amarthadan: You mean...like this? :smile: 
Screenshot_20220412_112500

src/make-request.ts
amarthadan: Fair enough, keep it.

src/make-request.ts
amarthadan: Do we actually need that distinction? I think this is the only part of the code that will deal with the `gateways` field. But I'm not against it, it just feels redundant.

src/validation.ts
amarthadan: > Beacons data must actually be bytes32 (string length 64) because of this https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/dapis/DapiServer.sol#L883

Didn't know that, that's fine then.

> for the signature length I am not sure, so I set it to regex+ (I don't think it matters much).

I think it's the same thing with the signature length. If it is in fact static we should enforce it (make the validation check) the same way we do for the value. We will avoid a potential unnecessary update call.

test/fixtures.ts
amarthadan: > That little?

https://github.com/api3dao/airnode/blob/master/packages/airnode-operation/src/server.ts This is 26 and contains logic we don't need so I think I'm pretty close :smile:

> Anyway, such a test will test a very limited happy path and the mock server can get out of sync with Airnode.

But that's true even for the Airnode itself. Even if we release a new Airnode the deployment has to be updated, same as the mock server would need to be updated.

> I'd rather treat this Airnode as an "external dependency" and go with this test even though there are these disadvantages. If this turns out to be a problem, we can remove this and implement the mock server. WDYT?

Let's do it the other way around. Let's start with the mock server (we're doing that for Airnode and Airkeeper at the moment anyway) and we can discuss and plan a proper deployment of a testing environment that will be used in all e2e tests across products. That would be actually neat, having like a real API provider and Airnode running we can make tests against. But at the moment I don't want a few tools using their own mock servers and one tool depending on an external dependency.

src/make-request.ts
Siegrift: Yeah, but in nodejs/chrome (v8) :D 

image

test/fixtures.ts
Siegrift: Ok, I'll revert.

src/make-request.ts
Siegrift: I'd create a low priority ticket for that, because the config.json schema should be updated as well, WDYT?

Siegrift: I will rename `makeSignedDataGatewayRequest` to `makeSignedDataGatewayRequests` but I still dislike that it receives an argument `gateway: Gateway` which is confusing imo.

src/validation.ts
Siegrift:
```
ECDSA signatures in Ethereum consist of three parameters: v , r and s . The signature is always 65-bytes in length. and returns the address used to sign the message.
```

Will set it to 130

src/make-request.ts
amarthadan: The fuck? :man_facepalming: 
Fair enough. But I still think this is easy enough problem so we don't need a library.
```ts
let url;
if (gatewayUrl.endsWith('/') {
  url = `${gatewayUrl}${endpointId}`;
} else {
  url = `${gatewayUrl}/${endpointId}`;
}
```
Or is there any other edge case?

src/make-request.ts
amarthadan: > I'd create a low priority ticket for that, because the config.json schema should be updated as well, WDYT?

Not really sure what ticket you want to create.

> I will rename makeSignedDataGatewayRequest to makeSignedDataGatewayRequests but I still dislike that it receives an argument gateway: Gateway which is confusing imo.

Kinda. As I said, I'm not against what you're proposing.

src/make-request.ts
Siegrift: I am not sure, but probably not. If you rally want it I can use if-else.

amarthadan: Yes, please do so :pray:

src/make-request.ts
Siegrift: Let me just do it now, we can drop it (the commit I'll create) if you disagree.

package.json
```diff
@@ -21,29 +21,37 @@
     "eslint:fix": "eslint . --ext .js,.ts --fix",
     "postinstall": "husky install",
     "pm2:logs": "pm2 logs",
-    "pm2:start": "scripts/run-locally.sh pm2 start ecosystem.config.js",
+    "pm2:start": "scripts/run-locally.sh pm2 start airseeker-service.config.js",
+    "pm2:start:testing-services": "pm2 start testing-services.config.js",
     "pm2:status": "pm2 status",
-    "pm2:stop": "pm2 delete ecosystem.config.js",
+    "pm2:stop": "pm2 delete airseeker-service.config.js",
+    "pm2:stop:testing-services": "pm2 delete testing-services.config.js",
```
amarthadan: I know that it's running pm2 but I would name it `dev:something` like we do in the rest. Also, please add separate targets for hardhat and mock server (I personally use those instead :sweat_smile:).

package.json
```diff
@@ -21,29 +21,37 @@
...
+    "promise.any": "^2.0.4",
     "zod": "^3.14.4"
   },
   "devDependencies": {
+    "@types/express": "^4.17.13",
     "@types/jest": "^27.4.1",
     "@types/lodash": "^4.14.181",
     "@types/node": "^17.0.23",
+    "@types/promise.any": "^2.0.0",
+    "@types/url-join": "^4.0.1",
```
amarthadan: You can remove this now, right?

test/server.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const app = express();
+
+app.post('/signed-data-gateway/endpoint', (_req: any, res: any) => {
+  res.status(200).send(validSignedData);
+});
+
+app.listen(PORT, () => {
+  console.log(`Server is running at http://localhost:${PORT}`);
+});
```
amarthadan, line 14: 14 lines and 4 of them empty. 10 exactly! Told you... :joy:

amarthadan (approved):
Just the two small things and this is good to go :+1:

test/server.ts
Siegrift: And x more dependencies and configuration files and package.json scripts :D 

Not even close :D

package.json
Siegrift: > Also, please add separate targets for hardhat and mock server (I personally use those instead 😅).

What targets?

amarthadan: https://github.com/api3dao/airkeeper/blob/psp-e2e-test/package.json#L26-L27

Siegrift: Actually I now realized that we do not need hardhat (at least as of now) so I'll remoeve that from the testing services config and only leave API there.

PR #26, opened 2022-04-13
Add deviation check

vponline:
This adds the functionality to check the DapiServer value for a **beaconId** and check if the updated API value deviation passes the deviationThreshold.


test/check-condition.feature.ts
```diff
@@ -0,0 +1,91 @@
+import fs from 'fs';
+import path from 'path';
+import { ethers } from 'ethers';
+import { checkUpdateCondition } from '../src/check-condition';
+
+const dapiServerAbi = JSON.parse(fs.readFileSync(path.resolve('./src/test/artifacts/DapiServer.json')).toString());
```
vponline: I used these tests to check that this actually worked with the Airkeeper DapiServer local PSP setup, and these could be used for E2E tests in the future. Otherwise I can delete if these are not needed.

src/check-condition.ts
```diff
@@ -0,0 +1,37 @@
...
+  const provider = new ethers.providers.JsonRpcProvider(providerUrl);
+  const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
+
+  // TODO readDataFeedValueWithId can throw if the dataFeed is not registered yet
+  // what do we want to return in this case
+  const dapiServerResponse = await dapiServer.connect(voidSigner).readDataFeedValueWithId(beaconId);
+  const updateInPercentage = calculateUpdateInPercentage(dapiServerResponse, ethers.BigNumber.from(updatedApiValue));
+
+  return updateInPercentage > deviationThreshold;
+};
```
vponline: Do we want a `try-catch` here for `readDataFeedValueWithId`?

bbenligiray: I assume you need to read the full data feed including the timestamp in Airseeker because it will need to update based on timestamp whatever the value is

vponline: :+1: Fixed. 

In this case, should I create a new function to read the value if it's needed in other places as well? Or will the timestamp logic be added here?

bbenligiray: We shouldn't do multiple calls to the blockchain provider to read the same value. @amarthadan can comment on how that should be implemented.

src/check-condition.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    expect(updateInPercentage).toEqual(80);
+  });
+});
+
+describe('checkUpdateCondition', () => {
+  const providerUrl = 'http://127.0.0.1:8545/';
+  const beaconId = '0x2ba0526238b0f2671b7981fd7a263730619c8e849a528088fd4a92350a8c2f2c';
+
+  let readDataFeedWithIdSpy: any;
+  let dapiServerMock: any;
```
acenolaza: Should we try to set proper types here?

src/check-condition.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const checkUpdateCondition = async (
+  providerUrl: string,
+  dapiServer: ethers.Contract,
+  beaconId: string,
+  deviationThreshold: number,
+  updatedApiValue: number
+) => {
+  const provider = new ethers.providers.JsonRpcProvider(providerUrl);
+  const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
```
acenolaza: These 2 seems like they should be initialized only once instead of doing it for every condition check.

src/check-condition.ts
```diff
@@ -0,0 +1,35 @@
...
+  // Convert back to a percentage number
+  return (change.toNumber() / HUNDRED_PERCENT) * 100;
+};
+
+export const checkUpdateCondition = async (
+  providerUrl: string,
+  dapiServer: ethers.Contract,
+  beaconId: string,
+  deviationThreshold: number,
+  updatedApiValue: number
```
acenolaza: Nit: just apiValue maybe? or mostUpToDateApiValue? or latestApiValue? or just keep current 😁

src/check-condition.ts
```diff
@@ -0,0 +1,35 @@
...
+) => {
+  const provider = new ethers.providers.JsonRpcProvider(providerUrl);
+  const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
+
+  const dapiServerResponse = await dapiServer.connect(voidSigner).readDataFeedWithId(beaconId);
```
acenolaza: Maybe here we might want to just return `true` if current datafeed timestamp value is older than x amount of time? WDYT @bbenligiray and skip the condition check?

acenolaza (commented):
A few comments but overall is looking good 👍🏻

src/check-condition.ts
```diff
@@ -0,0 +1,35 @@
...
+  const delta = updatedValue.sub(initialValue);
+  const absoluteDelta = delta.abs();
+
+  // Avoid division by 0
+  const absoluteInitialValue = initialValue.isZero() ? ethers.BigNumber.from(1) : initialValue.abs();
+
+  const change = absoluteDelta.mul(ethers.BigNumber.from(HUNDRED_PERCENT)).div(absoluteInitialValue);
+
+  // Convert back to a percentage number
+  return (change.toNumber() / HUNDRED_PERCENT) * 100;
```
acenolaza: I have a really hard time trusting integers in js and I wonder if we should just keep doing the math using BigNumber instead 🤔

vponline: That makes sense to me now that you mentioned :smile:

src/check-condition.ts
amarthadan: @bbenligiray Do you mean if different sponsors (on the same chain) want to read the same beacon? I'm not sure how that can be easily done since you can configure different update intervals for each sponsor. Or am I missing something?

Or do you mean the same chain, the same sponsor, and multiple providers?

src/check-condition.ts
amarthadan: Yes, what's @acenolaza saying is true. You can expect the provider (an initialized provider object) to be passed to the function.

src/check-condition.ts
amarthadan: That's going to be added as part of [BEC-297](https://api3dao.atlassian.net/browse/BEC-297).

src/check-condition.ts
vponline: Then we could just pass in the voidSigner right? We don't need the provider itself.

src/check-condition.test.ts
```diff
@@ -0,0 +1,117 @@
+import { ethers } from 'ethers';
+import { calculateUpdateInPercentage, checkUpdateCondition, HUNDRED_PERCENT } from './check-condition';
+
+describe('calculateUpdateInPercentage', () => {
```
Siegrift, line 4: I don't think we need so many tests. I'd drop the `calculates increase` and  `calculates decrease` since they are implicitely tested by the other ones.

src/check-condition.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
+
+  let readDataFeedWithIdSpy: jest.Mock;
+  let dapiServerMock: {
+    connect(_signerOrProvider: ethers.Signer | ethers.providers.Provider | string): typeof dapiServerMock;
+    readDataFeedWithId: jest.Mock;
+  };
+
+  beforeEach(() => {
+    const readDataFeedWithIdMock = (_beaconId: string) => Promise.resolve([ethers.BigNumber.from(500)]);
```
Siegrift: Nit: you don't need to specify arguments if you don't need to use them.

src/check-condition.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  it('reads dapiserver value and checks the threshold condition to be true for decrease', async () => {
+    const readDataFeedWithIdOnceSpy = jest
+      .fn()
+      .mockImplementationOnce(() => Promise.resolve([ethers.BigNumber.from(400)]));
+    const checkUpdateConditionResult = await checkUpdateCondition(
+      voidSigner,
+      {
+        ...dapiServerMock,
```
Siegrift: Why can't you do the same mocking as in the beforeEach hook? I think it would be a bit cleaner not to use the beforeEach and mock `dapiServerMock` in each of the tests.

src/check-condition.ts
```diff
@@ -0,0 +1,30 @@
...
+// `calculateUpdateInPercentage()`. Since the reported data needs to fit
+// into 224 bits, its multiplication by 10^8 is guaranteed not to overflow.
+export const HUNDRED_PERCENT = 1e8;
+
+export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, updatedValue: ethers.BigNumber) => {
+  const delta = updatedValue.sub(initialValue);
+  const absoluteDelta = delta.abs();
+
+  // Avoid division by 0
+  const absoluteInitialValue = initialValue.isZero() ? ethers.BigNumber.from(1) : initialValue.abs();
```
Siegrift, line 13: Mhh, I wonder if this edge case should be explained somewhere (for the API integrators).

src/check-condition.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const checkUpdateCondition = async (
+  voidSigner: ethers.VoidSigner,
+  dapiServer: ethers.Contract,
+  beaconId: string,
+  deviationThreshold: number,
+  apiValue: number
+) => {
+  const [dapiServerValue, _timestamp] = await dapiServer.connect(voidSigner).readDataFeedWithId(beaconId);
```
Siegrift, line 25: We also want to to update if the response is too old (not sure if that is part of this task though)

test/check-condition.feature.ts
```diff
@@ -0,0 +1,83 @@
+import { ethers } from 'ethers';
+import { DapiServer__factory as DapiServerFactory } from '@api3/airnode-protocol-v1';
+import { checkUpdateCondition } from '../src/check-condition';
+
+describe('checkUpdateCondition', () => {
+  const providerUrl = 'http://127.0.0.1:8545/';
+  const provider = new ethers.providers.JsonRpcProvider(providerUrl);
+  const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
+  const dapiServer = new ethers.Contract('0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0', DapiServerFactory.abi, provider);
```
Siegrift: 1) I am missing scripts to deploy this contract
2) This test is not seem to be recognized by jest configuration.
3) Can you rename the `e2e.test.ts` to `make-request.feature.ts`? Also, maybe put them into folder `e2e`.

Siegrift (changes requested):

test/check-condition.feature.ts
vponline: I will cover all of these in the Airseeker E2E task.

(if you wanted to run these you could deploy the Airkeeper local PSP + run the update script).

src/check-condition.ts
vponline: Will be covered in another task :smile:

src/check-condition.ts
vponline: What kind of case were you thinking? (I thought this is just a math limitation)

acenolaza (approved):

src/check-condition.ts
Siegrift: > (I thought this is just a math limitation)

It is, but we have to handle it somehow and using 1 seems the best we could do.

> What kind of case were you thinking?

If the value was `0` and now is `5` it doesn't make sense to ask "how big is the deviation" (because there is the division by 0). We divide by 1 in that case, which seems reasonable but it behaves strangely since deviation from 0 -> 1 leads the same percentage update as 1 -> 2.

test/check-condition.feature.ts
Siegrift: I think at least this test should be included in the pipeline before the PR is merged.

Siegrift (commented):
👍 The functionality LGTM, but I think the e2e test should be included in the pipeline.

test/check-condition.feature.ts
vponline: Ok, I can add the setup for these from the E2E branch :smile:

src/check-condition.ts
vponline: > We divide by 1 in that case, which seems reasonable but it behaves strangely since deviation from 0 -> 1 leads the same percentage update as 1 -> 2.

Shouldn't this be fine as the result is that the beacon should be updated? The actual update percentage doesn't really matter in this case (when the **dapiServerValue** is 0), right? Perhaps @amarthadan can comment as well.

(For context, this function is essentially the same as the smart contract function [here](https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol#L862))

Siegrift: Yeah, the value is 0 initially, so after the first beacon update should always update the value on chain. 
However, the value *could* be also set to 0 by the API and it's not clear whether the next non zero value should also trigger update. But this is an edge case, which I can't imagine happening in practice, so I think the implementation is good.

test/config/config.ts
```diff
@@ -0,0 +1,131 @@
+export const buildAirseekerConfig = () => ({
```
Siegrift: It would be better to move this to `test/fixtures`

test/setup/deployment.ts
```diff
@@ -0,0 +1,231 @@
+import { Contract, Wallet } from 'ethers';
+import * as hre from 'hardhat';
+import * as abi from '@api3/airnode-abi';
+// TODO uncomment once airnode-node 0.6 is released
```
Siegrift: Nit: (more like my ocd :D)

```
TODO: uncomment...
```

tsconfig.json
```diff
@@ -10,8 +10,8 @@
     "target": "es5",
     "skipLibCheck": true,
     "baseUrl": "./",
-    "outDir": "./dist/src"
+    "outDir": "./dist"
   },
-  "include": ["./src/**/*.ts"],
+  "include": ["./src/**/*.ts", "./test/**/*", "./hardhat.config.ts"],
```
Siegrift: This is OK for now, but we should have two separate tsconfigs (one for development and one for prod), but this is tracked in https://api3dao.atlassian.net/browse/BEC-311

Siegrift (approved):
👍 LGTM

.github/workflows/main.yml
```diff
@@ -48,6 +48,26 @@ jobs:
...
+          cache: 'yarn'
+      - name: Install Dependencies
+        run: yarn install --frozen-lockfile
+      - name: Build
+        run: yarn build
+      - name: Start background services
+        run: yarn dev:testing-services:start
+      - name: Test
+        run: yarn test:e2e
+
```
amarthadan: Is this actually faster than having it run after the unit tests? If so, I'm ok with it but you need to add the `if` key to stop tests from running twice for PRs and you need to remove `yarn dev:testing-services:start` from the `lint-build-test` job above.

package.json
```diff
@@ -12,8 +12,10 @@
     "build": "yarn clean && yarn compile",
     "clean": "rimraf -rf ./dist *.tgz",
     "compile": "tsc -p tsconfig.json",
-    "dev:api": "ts-node test/server.ts",
-    "dev:testing-services:start": "pm2 start testing-services.config.js",
+    "dev:clean": "rimraf -rf ./logs",
```
amarthadan: Why do we even create the `./logs` directory? If it's just for tests can't we log either to STDOUT or `/dev/null` or turn it off altogether if we don't need it?

src/check-condition.ts
amarthadan: I agree with both of you. I think we should follow the implementation from the contract and I agree that this implementation is fine.

tsconfig.json
amarthadan: Why do we need this? Tests are actually excluded in https://github.com/api3dao/airseeker/pull/28 to make the build process work correctly.

amarthadan (approved):
I'm approving this but please take a look at my comments before merging.

package.json
vponline: These were removed from Airkeeper as well so I'll remove

.github/workflows/main.yml
vponline: I dont believe it is faster, I'll change it to run after the unit tests (like you had done in beacon-updates)

PR #28, opened 2022-04-15
BEC-315:Fix where's the configuration loaded from in the container

acenolaza:
I ended up moving the src folder out of dist folder inside the container to match the path of `__dirname` value when running Airseeker locally using `yarn start`

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,6 +12,6 @@
     "baseUrl": "./",
     "outDir": "./dist"
   },
-  "include": ["./src/**/*.ts", "./hardhat.config.ts"],
-  "exclude": ["node_modules"]
+  "include": ["./src/**/*.ts"],
+  "exclude": ["node_modules", "src/**/*.test.ts"]
 }
--- a/airseeker-service.config.js
+++ b/airseeker-service.config.js
@@ -3,7 +3,7 @@ module.exports = {
   apps: [
     {
       name: 'airseeker',
-      script: './src/main.js',
+      script: './dist/main.js',
       kill_timeout: 10_000,
       env: {
         NODE_ENV: 'development',
--- a/airseeker-service.config.js
+++ b/airseeker-service.config.js
@@ -3,7 +3,7 @@ module.exports = {
   apps: [
     {
       name: 'airseeker',
-      script: './dist/src/main.js',
+      script: './src/main.js',
       kill_timeout: 10_000,
       env: {
         NODE_ENV: 'development',
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,7 +40,9 @@ RUN yarn install && \
     yarn build && \
     yarn pack && \
     mkdir -p ${packageDir} && \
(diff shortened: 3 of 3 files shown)


tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "target": "es5",
     "skipLibCheck": true,
     "baseUrl": "./",
-    "outDir": "./dist/src"
+    "outDir": "./dist"
```
acenolaza, line 13: This is to fix this issue: https://api3workspace.slack.com/archives/C037NLX1R8T/p1649954030733229

amarthadan (changes requested):
This won't work because you won't be able to run pm2 locally.

The ideal solution would be to build everything directly into a `dist` directory. That way we can point everything to `dist/main.js` and the program will look for the config file in `${__dirname}/../config/airseeker.json` and will find it.

Problem is that currently when you run `yarn build` the `dist` directory contains `src` directory (instead source files directly) and `test` directory. I don't really know why the `test` directory is included but once removed I think it will build directly into `dist` which will solve the problem.

acenolaza:
> This won't work because you won't be able to run pm2 locally.

Right, I only tested `yarn start` script when testing locally

> The ideal solution would be to build everything directly into a `dist` directory. That way we can point everything to `dist/main.js` and the program will look for the config file in `${__dirname}/../config/airseeker.json` and will find it.
> 
> Problem is that currently when you run `yarn build` the `dist` directory contains `src` directory (instead source files directly) and `test` directory. I don't really know why the `test` directory is included but once removed I think it will build directly into `dist` which will solve the problem.

Thanks for the feedback. Let me try this approach then 👍🏻

tsconfig.json
```diff
@@ -10,8 +10,8 @@
...
   },
   "include": ["./src/**/*.ts"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules", "src/**/*.test.ts", "dist/**/*"]
```
amarthadan: Interesting. I didn't think about removing test files from `src`. I guess it was including the `test` directory because some tests (in `src`) load fixtures from there. But I think you should be able to remove the `"dist/**/*"` part.

amarthadan (approved):
LGTM :+1:

tsconfig.json
acenolaza: Correct. There were tests importing fixtures causing the folder structure (`src` and `test`) to be created inside `dist`. I can remove the `"dist/**/*"` part but I just wanted to be consistent with Airkeeper build [tsconfig.json](https://github.com/api3dao/airkeeper/blob/main/tsconfig-build.json) file. Should I still remove it?

amarthadan: Feel free to leave it there but I don't think it does anything since the `outDir` is set to `./dist` and the whole `dist` directory is always removed before the build :man_shrugging:

acenolaza: Then I guess we should remove it in both places 😁

PR #29, opened 2022-04-16
BEC-317 | Set Gateway URL as secret in example config

Ashar2shahid:
I don't think this needs changes anywhere else afaik


config/secrets.example.env
```diff
@@ -3,3 +3,4 @@ CP_INFURA_MAINNET_URL=""
 CP_INFURA_ROPSTEN_URL=""
 
 HTTP_GATEWAY_API_KEY=""
+HTTP_GATEWAY_API_URL=""
```
amarthadan: Can you call it `HTTP_GATEWAY_URL` instead? API key makes sense but API URL sounds weird :smile:

amarthadan (approved):

config/secrets.example.env
Ashar2shahid: We will be using [this ](https://github.com/api3dao/operations/pull/26/files#diff-c6d8c75d3020dd8a2b52339ee1cae7bf411266bf7b22c1f56d3b353a4cc3263cR165)naming scheme in operations, so I will change it according to that (Without the AWS)

PR #33, opened 2022-04-19
Beacon updates

amarthadan:
[BEC-249](https://api3dao.atlassian.net/browse/BEC-249)

I'm opening this as a draft to get some early reviews and unblock other tasks. I'm still missing tests and potentially resolving some TODOs but the code itself should work. I was able to test it by deploying and setting wallets with Aireekeper and using the mock server for the HTTP gateway.

The PR is a bit bigger but it should be fine when reviewing commit by commit :slightly_smiling_face:


src/state.ts
```diff
@@ -1,19 +1,32 @@
...
 export type BeaconValueStorage = Record<BeaconId, SignedData>;
+export type Provider = {
+  rpcProvider: ethers.providers.StaticJsonRpcProvider;
+  chainId: string;
+};
+// chainId => Provider[]
+export type Providers = Record<string, Provider[]>;
 
 export interface State {
+  config: Config;
```
amarthadan, line 13: Moving config to the state seemed like a good idea to me, it's quite handy that I don't need to pass it everywhere. But I'm happy to revert it if you don't like it.

src/update-beacons.ts
```diff
@@ -0,0 +1,204 @@
...
+    const waitTime = Math.max(0, updateInterval * 1_000 - duration);
+    await sleep(waitTime);
+  }
+};
+
+const calculateCurrentTimeout = (startTime: number, totalTimeout: number) => totalTimeout - (Date.now() - startTime);
+
+// We retry all chain operations with a random back-off infinitely until the next updates cycle
+// TODO: Errors are not displayed with this approach. Problem?
+const getCurrentProviderGoOptions = (startTime: number, totalTimeout: number): GoAsyncOptions => ({
```
amarthadan: I'm passing this to all the chain operations to kinda enforce the update cycle. Since chain operations are the only ones that actually take some time I think this is a good enough solution. But I'm open to suggestions.

src/utils.ts
```diff
@@ -1 +1,22 @@
+import { ethers } from 'ethers';
+
 export const sleep = (ms: number) => new Promise((res) => setTimeout(res, ms));
+
+export const shortenAddress = (address: string) => address.replace(address.substring(5, 38), '...');
+
+// TODO: Remove once airnode-node 0.6.0 is released
```
amarthadan, line 7: I didn't want to create another crazy build-airnode-first dependency for Airseeker so I just copy&pasted it from Airnode. Will be removed.

src/main.ts
```diff
@@ -23,6 +24,8 @@ const updateBeacons = async () => {
   }
 };
 
+initializeProviders(config);
+
```
Siegrift, line 19: Nit: Why \n?

src/providers.ts
```diff
@@ -0,0 +1,40 @@
+import * as node from '@api3/airnode-node';
+import {DapiServer__factory} from '@api3/airnode-protocol-v1';
```
Siegrift: This looks incorrectly formatted?

Siegrift: Also consider importing `DapiServer__factory as DapiServerFactory`

src/state.ts
```diff
@@ -1,15 +1,26 @@
...
+import { DapiServer } from '@api3/airnode-protocol-v1';
 import { BeaconId, SignedData } from './validation';
 
 export type BeaconValueStorage = Record<BeaconId, SignedData>;
+export type Provider = {
+  contract: DapiServer;
+  rpcProvider: ethers.providers.StaticJsonRpcProvider;
+  chainId: string;
+};
+// chainId => Provider[]
```
Siegrift, line 9: ?

src/validation.ts
```diff
@@ -22,19 +23,8 @@ export const beaconSetsSchema = emptyObjectSchema;
...
-    ),
-    options: z.object({
-      txType: z.string(),
-      priorityFee: z.object({
-        value: z.number(),
-        unit: z.string(),
-      }),
-      baseFeeMultiplier: z.number(),
-    }),
+    providers: z.record(providerSchema),
```
Siegrift, line 26: Do we want to couple this with validation from Airnode? (I think it's OK, but we could do this in a different task and across other repos as well).

src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const INFINITE_RETRIES = 100_000;
+export const GATEWAY_TIMEOUT_MS = 5_000;
+export const PROVIDER_TIMEOUT_MS = 5_000;
```
Siegrift, line 5: Where is this used?

src/state.ts
Siegrift: It will make testing a bit harder (since all of the "pure" functions now depend on a "global" state, but I think we can keep it.

src/gas-prices.ts
```diff
@@ -0,0 +1,89 @@
...
+  let gasTargetMessage;
+  if (gasTarget.txType === 'eip1559') {
+    const gweiMaxFee = node.evm.weiToGwei(gasTarget.maxFeePerGas!);
+    const gweiPriorityFee = node.evm.weiToGwei(gasTarget.maxPriorityFeePerGas!);
+    gasTargetMessage = `Gas price (EIP-1559) set to a Max Fee of ${gweiMaxFee} Gwei and a Priority Fee of ${gweiPriorityFee} Gwei`;
+  } else {
+    const gweiPrice = node.evm.weiToGwei(gasTarget.gasPrice!);
+    gasTargetMessage = `Gas price (legacy) set to ${gweiPrice} Gwei`;
+  }
+  console.log(gasTargetMessage);
```
Siegrift, line 86: rm?

src/update-beacons.ts
Siegrift: I think it's fine, but this should be commented on the `updateBeacons` function.

Siegrift: Also maybe just call this `getGoOptions` to make it shorter.

src/update-beacons.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    await updateBeacons(providerSponsorBeacons);
+
+    const duration = Date.now() - startTimestamp;
+    const waitTime = Math.max(0, updateInterval * 1_000 - duration);
+    await sleep(waitTime);
+  }
+};
+
+const calculateCurrentTimeout = (startTime: number, totalTimeout: number) => totalTimeout - (Date.now() - startTime);
```
Siegrift: Maybe `getTimeout`?

Siegrift (commented):
Did a quick pass and left a few comments.

src/block-number.ts
```diff
@@ -0,0 +1,16 @@
+import { go, GoAsyncOptions } from '@api3/promise-utils';
+import { Provider } from './state';
+
+export const getCurrentBlockNumber = async (provider: Provider, goOptions: GoAsyncOptions): Promise<number | null> => {
+  const goBlockNumber = await go(() => provider.rpcProvider.getBlockNumber(), goOptions);
+
+  if (!goBlockNumber.success) {
+    console.log(`Unable to get block number. Error: ${goBlockNumber.error}`);
```
Siegrift, line 8: We are probably overlogging as this will also trigger similar console.log in `updateBeacons`. 

It would be nice to have different log levels and have these logs with lower severity then the ones in `updateBeacons`.

src/check-condition.ts
```diff
@@ -0,0 +1,11 @@
+import { ethers } from 'ethers';
+
+export const checkUpdateCondition = async (
+  _voidSigner: ethers.VoidSigner,
+  _dapiServer: ethers.Contract,
+  _beaconId: string,
+  _deviationThreshold: number,
+  _apiValue: ethers.BigNumber
+) => {
+  return true;
```
Siegrift: TODO:

src/constants.ts
Siegrift: Ok, I see now, but I think it was not in the commit in which it was added.

src/gas-prices.ts
```diff
@@ -0,0 +1,89 @@
+// Taken from @api3/airnode-utils so custom retry & timeout options can be used
```
Siegrift, line 1: I don't understand. Which custom retry & timeout options do you mean?

src/update-beacons.ts
```diff
@@ -0,0 +1,224 @@
...
+  GAS_LIMIT,
+  INFINITE_RETRIES,
+  PROTOCOL_ID,
+  PROVIDER_TIMEOUT_MS,
+  RANDOM_BACKOFF_MAX_MS,
+  RANDOM_BACKOFF_MIN_MS,
+} from './constants';
+
+// Solidity type(int224).min
+const INT224_MIN = ethers.BigNumber.from(2).pow(ethers.BigNumber.from(223)).mul(ethers.BigNumber.from(-1));
```
Siegrift: Move to constants?

src/update-beacons.ts
```diff
@@ -0,0 +1,224 @@
...
+    );
+    if (newBeaconValue.gt(INT224_MAX) || newBeaconValue.lt(INT224_MIN)) {
+      console.log(`New beacon value for beacon with ID ${beaconUpdateData.beaconId} is out of type range. Skipping.`);
+      continue;
+    }
+
+    // Check beacon condition
+    // TODO: Add retry and rest of the go options
+    //       Will finish once https://github.com/api3dao/airseeker/pull/26 is merged
+    const shouldUpdate = await checkUpdateCondition(
```
Siegrift, line 196: I think it would be better to check these update conditions in parallel and only do the final submissions sequentially. WDYT?

src/update-beacons.ts
```diff
@@ -0,0 +1,224 @@
...
+// We retry all chain operations with a random back-off infinitely until the next updates cycle
+// TODO: Errors are not displayed with this approach. Problem?
+const getCurrentProviderGoOptions = (startTime: number, totalTimeout: number): GoAsyncOptions => ({
+  attemptTimeoutMs: PROVIDER_TIMEOUT_MS,
+  totalTimeoutMs: calculateCurrentTimeout(startTime, totalTimeout),
+  retries: INFINITE_RETRIES,
+  delay: { type: 'random' as const, minDelayMs: RANDOM_BACKOFF_MIN_MS, maxDelayMs: RANDOM_BACKOFF_MAX_MS },
+});
+
+export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorBeacons) => {
```
Siegrift, line 104: Maybe split this function into fewer smaller parts? (e.g. submit beacon update)

src/gas-prices.ts
Siegrift: I think you meant custom prio and base fee?

src/main.ts
amarthadan: To divide initialization and actual processing.

src/state.ts
amarthadan: That's the structure explanation. The key is the chain ID and the value is an array of `Provider` objects.

src/validation.ts
amarthadan: I ran into this when working with `chains.options` portion of the config as we originally didn't list all the unit types for fees. The other option was to copy & paste it from Airnode. As for other repos, Airkeeper is currently loading the whole Airnode config and using the whole Airnode validation. Don't know about ops repo for example.

src/gas-prices.ts
amarthadan: No, it just looks that way because it's a different message for legacy and for EIP1559 pricing.

src/update-beacons.ts
amarthadan: I wanted to emphasize that the value is calculated and therefore different each time the function is called instead of just retrieved from somewhere.

src/block-number.ts
amarthadan: I agree, that can be part of [BEC-336](https://api3dao.atlassian.net/browse/BEC-336).

src/gas-prices.ts
amarthadan: No, I meant retry & timeout options. If I would just import the function from `airnode-utils` I would not have control over those options. It would try to obtain the gas price just three times or something like that (not sure if with a random back-off) and then stop. This makes sense for the lambda environment where Airnode's run but it doesn't for the Airseeker use case which runs continuously. We can't move without the gas price so we may as well try to fetch it until the total timeout (the update cycle) anyway.

src/update-beacons.ts
amarthadan: I think the point of doing this sequentially was not to overwhelm the blockchain RPC provider. But we can discuss it on Slack.

src/update-beacons.ts
amarthadan: Yeah, I was thinking about this as well.

src/make-request.ts
```diff
@@ -28,7 +28,7 @@ export const makeSignedDataGatewayRequests = async (
         method: 'POST',
         headers: { 'x-api-key': apiKey },
         data: JSON.stringify({ encodedParameters: parameters }),
```
vponline: While working on the E2E tests, I had to remove the **JSON.stringify** from `data` or add a **Content-Type** `header`.

It doesn't look like data is stringified in Airnode's requests either (e.g. [here](https://github.com/api3dao/airnode/blob/master/packages/airnode-adapter/src/request-building/execution.test.ts)). 

So just wondering if this was on purpose. (I realize it's not from this PR)

amarthadan: Yeah, it might be that the `axios` stringifies it for you. But we definitely should add (or make sure `axios` is adding it) content type saying that this is a JSON and also that we're expecting JSON back. I think AWS gateway would work ok without it but if I remember correctly the GCP one requires it :confused: 
I've opened [BEC-340](https://api3dao.atlassian.net/browse/BEC-340) for this.

Siegrift: Yeah, I tested it with AWS and it worked there, but adding the content type header is a a good idea.

src/check-condition.ts
amarthadan: It's already merged so I'll rebase and connect it properly.

src/block-number.ts
Siegrift: Since I'm doing 336 now, I'd maybe defer that later. Also, it's not clear to me which messages should have which log level. I think at least for starters it is OK if we are overlogging as it will make it easier to debug... 

I think we can address the log levels issue after our service is stable and we identify "spammy" messages which are not that useful. WDYT?

src/gas-prices.ts
acenolaza: I might be wrong but `@api3/airnode-utilities.getGasPrice()` receives a `FetchOptions` type arg which contains a variable of type `ChainOptions` which extends `PromiseOptions` type. Meaning that `retries`, `retryDelayMs` and `timeoutMs` should be available to the caller of `@api3/airnode-utilities.getGasPrice()`.

src/state.ts
acenolaza: Nit: Do we really need the `Provider` type? The `chainId` seems a bit redundant. Maybe we can just create a `Record<string, ethers.providers.StaticJsonRpcProvider[]>`

src/gas-prices.ts
amarthadan: Yes, it is but it's using a custom `go` implementation using the `attempt` library not the one from our `promise-utils` library that supports proper timeout and random back-off.

amarthadan:
Ok, this is ready to be reviewed now @acenolaza @vponline @Siegrift @aquarat 
I'll add tests in a separate PR, this one is big enough as it is.

src/update-beacons.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    return [...acc, ...providerSponsorGroups];
+  }, []);
+};
+
+export const initiateBeaconUpdates = () => {
+  console.log('Initiating beacon updates');
+
+  const providerSponsorBeaconsGroups = groupBeaconsByProviderSponsor();
+  providerSponsorBeaconsGroups.forEach(updateBeaconsInLoop);
```
acenolaza: Should we check if `providerSponsorBeaconsGroups` is empty and log a warning message if so?

src/update-beacons.ts
```diff
@@ -0,0 +1,227 @@
...
+    // TODO: Should be later part of the validation
+    if (!providers) {
+      console.log(`Missing providers for chain with ID ${chainId}`);
+      return acc;
+    }
+
+    const providerSponsorGroups = Object.entries(sponsors).reduce(
+      (acc: ProviderSponsorBeacons[], [sponsorAddress, beaconUpdate]) => {
+        const { beacons, updateInterval } = beaconUpdate;
+        return [...acc, ...providers.map((provider) => ({ provider, sponsorAddress, updateInterval, beacons }))];
```
acenolaza: Nit: 

```suggestion
        return [...acc, ...providers.map((provider) => ({ provider, sponsorAddress, ...beaconUpdate }))];
```

src/update-beacons.ts
```diff
@@ -0,0 +1,227 @@
...
+};
+
+export const updateBeaconsInLoop = async (providerSponsorBeacons: ProviderSponsorBeacons) => {
+  while (!getState().stopSignalReceived) {
+    const startTimestamp = Date.now();
+    const { updateInterval } = providerSponsorBeacons;
+
+    await updateBeacons(providerSponsorBeacons);
+
+    const duration = Date.now() - startTimestamp;
```
acenolaza, line 85: Do we want to console.log the `duration` like in Airnode and Airkeeper?

src/update-beacons.ts
```diff
@@ -0,0 +1,227 @@
...
+  const contractAddress = config.chains[chainId].contracts['DapiServer'];
+  // TODO: Should be later part of the validation
+  if (!contractAddress) {
+    console.log(`Missing contract address for DapiServer on chain with ID ${chainId}.`);
+    return;
+  }
+  const contract = DapiServerFactory.connect(contractAddress, rpcProvider);
+
+  // Get current block number
+  const blockNumber = await getCurrentBlockNumber(provider, getGoOptions(startTime, totalTimeout));
```
acenolaza: Nit: Maybe we can put the result of calling `getGoOptions(...)` in a const after line 95 since we are calling this function multiple times with the same args within this function.

src/update-beacons.ts
```diff
@@ -0,0 +1,227 @@
...
+  const providerSponsorBeaconsGroups = groupBeaconsByProviderSponsor();
+  providerSponsorBeaconsGroups.forEach(updateBeaconsInLoop);
+};
+
+export const updateBeaconsInLoop = async (providerSponsorBeacons: ProviderSponsorBeacons) => {
+  while (!getState().stopSignalReceived) {
+    const startTimestamp = Date.now();
+    const { updateInterval } = providerSponsorBeacons;
+
+    await updateBeacons(providerSponsorBeacons);
```
acenolaza, line 83: Maybe this has been already discussed but shouldn't we process all providers in parallel?

src/update-beacons.ts
```diff
@@ -0,0 +1,227 @@
...
+    const shouldUpdate = await checkUpdateCondition(
+      voidSigner,
+      contract,
+      beaconUpdateData.beaconId,
+      beaconUpdateData.deviationThreshold,
+      newBeaconValue,
+      getGoOptions(startTime, totalTimeout)
+    );
+    if (shouldUpdate === null) {
+      console.log(`Unable to fetch current beacon value for beacon with ID ${beaconUpdateData.beaconId}.`);
```
acenolaza, line 205: This message makes me think that this function knows too much details about the `checkUpdateCondition` function internals...I wonder if we should take the fetching of the beacon value out of that function and feed it as an arg instead 🤔

acenolaza (approved):
This is looking really good 🚀

src/update-beacons.ts
amarthadan: Nice suggestion. It prompted me to add additional configuration checks for both beacon updates and data fetching :+1:

src/update-beacons.ts
amarthadan: Not entirely sure this would be useful for Airseeker since it's running continuously instead of in lambda like Airkeeper. Maybe some statistic about how many beacons we were able to process within the update interval would be more useful.

src/update-beacons.ts
amarthadan: We can't do that because the total timeout needs to be calculated every time. I'll change the name of the function so it's clear.

src/update-beacons.ts
amarthadan: We are. We're processing beacons for provider + sponsor pair in parallel. What you see here is one provider + sponsor pair and its beacons. The parallelization is happening in `initiateBeaconUpdates`.

src/update-beacons.ts
amarthadan: You are totally right. And we will have to most likely do it for [BEC-341](https://api3dao.atlassian.net/browse/BEC-341) and [BEC-297](https://api3dao.atlassian.net/browse/BEC-297) so I'm leaving it for those issues.

src/fetch-beacon-data.ts
```diff
@@ -1,24 +1,43 @@
...
+      });
     })
   );
 
-  beaconIdsToUpdate.forEach((id) => {
-    fetchBeaconDataInLoop(config, id);
-  });
+  if (isEmpty(beaconIdsToUpdate)) {
+    console.log('No beacons to fetch data for found. Stopping.');
+    process.exit(NO_FETCH_EXIT_CODE);
```
Siegrift, line 37: If we have this in place we should remove the stopSignal in `state` and just process.exit when immediatelly (not in this task though).

Siegrift (approved):
👍 LGTM

src/fetch-beacon-data.ts
amarthadan: I agree. The current graceful waiting is in practice quite annoying and it actually brings nothing :sweat_smile: That's on me.

src/fetch-beacon-data.ts
```diff
@@ -1,24 +1,43 @@
...
+  INFINITE_RETRIES,
+  NO_FETCH_EXIT_CODE,
+  RANDOM_BACKOFF_MAX_MS,
+  RANDOM_BACKOFF_MIN_MS,
+} from './constants';
 
-export const initiateFetchingBeaconData = async (config: Config) => {
+export const initiateFetchingBeaconData = async () => {
   console.log('Initiating fetching all beacon data');
+  const config = getState().config;
```
vponline:
```suggestion
  const { config } = getState();
```
:detective: Not significant, but saw this in a couple of places.

vponline (approved):

PR #34, opened 2022-04-20
Use Airnode logging utilities

Siegrift:
https://api3dao.atlassian.net/browse/BEC-336

Use Airnode logging utilities. For now I've just replaced the console.log calls with a logger with hard-coded configuration which prints timestamp and log level alongside with the message.

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -61,7 +61,7 @@ module.exports = {
     // ESLint
     'comma-dangle': ['error', 'only-multiline'],
     indent: 'off',
-    'no-console': 'off',
+    'no-console': 'error',
     'no-useless-escape': 'off',
     semi: 'error',
     eqeqeq: ['error', 'smart'],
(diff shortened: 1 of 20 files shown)


amarthadan:
@Siegrift Can you open this against `beacon-updates` branch and switch all the logging there as well?

.eslintrc.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
     // ESLint
     'comma-dangle': ['error', 'only-multiline'],
     indent: 'off',
-    'no-console': 'off',
+    'no-console': 'error',
```
Siegrift, line 64: I set this to `error` as we should be using `logger` now. This is slightly awkward in tests, but we should not use console.log there either (and we can disable with inline comments)

Siegrift:
> Can you open this against beacon-updates branch and switch all the logging there as well?

Sure, but what about other open PRs? I think we can just wait until they are merged.

amarthadan:
> > Can you open this against beacon-updates branch and switch all the logging there as well?
> 
> Sure, but what about other open PRs? I think we can just wait until they are merged.

The Terraform one won't have much code inside if any and the configuration file placement one has just some changes in TS configuration. The beacon updates one is the only one with actual code and a lot of log statements that's why I'm asking for it.

vponline (approved):
I think it looks good to go :+1:

PR #31, opened 2022-04-18
Add E2E tests

vponline:
This adds E2E tests for Airseeker.

PR #36, opened 2022-04-22
Unit & E2E tests for beacon updates

amarthadan:
I know that the E2E test is a weak one but it took me a stupid amount of time to make sure all the IDs are in sync and I want to move on. We have https://github.com/api3dao/airseeker/pull/31 anyway and we can add more tests later.

I had to add `--runInBand` for E2E tests so they run sequentially because otherwise the chain resets messed up unrelated tests.

### DIFF
--- a/src/gas-prices.test.ts
+++ b/src/gas-prices.test.ts
@@ -1,0 +1,166 @@
+import { BigNumber, ethers } from 'ethers';
+import { PriorityFee } from '@api3/airnode-node';
+import * as state from './state';
+import * as gasPrices from './gas-prices';
+import { BASE_FEE_MULTIPLIER, PRIORITY_FEE_IN_WEI } from './constants';
+
+const setupMocks = () => {
+  jest.spyOn(ethers.providers, 'JsonRpcProvider').mockImplementation(
+    () =>
+      ({
+        getGasPrice: jest.fn(),
+        getBlock: jest.fn(),
+      } as unknown as ethers.providers.StaticJsonRpcProvider)
+  );
+  jest.spyOn(state, 'getState').mockImplementation(createMockState);
+};
+
+const createProvider = (chainId: string) => ({
+  rpcProvider: new ethers.providers.JsonRpcProvider(),
+  chainId,
+});
+
+const createMockState = () =>
+  ({
+    config: {
+      chains: {
+        '31337': {
+          options: {
+            txType: 'legacy',
+          },
+        },
+        '31331': {
+          options: {
+            txType: 'eip1559',
+            baseFeeMultiplier: BASE_FEE_MULTIPLIER,
+            priorityFee: {
+              value: 3.12,
+              unit: 'gwei',
+            },
+          },
+        },
+        '31332': {
+          options: {
+            txType: 'eip1559',
+            baseFeeMultiplier: undefined,
+            priorityFee: {
+              value: 3.12,
+              unit: 'gwei',
+            },
+          },
+        },
+        '31333': {
+          options: {
+            txType: 'eip1559',
+            baseFeeMultiplier: BASE_FEE_MULTIPLIER,
+            priorityFee: undefined,
+          },
+        },
+        '31334': {
+          options: {
+            txType: 'eip1559',
+            baseFeeMultiplier: undefined,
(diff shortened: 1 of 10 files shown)


vponline (approved):
Nice

src/gas-prices.test.ts
```diff
@@ -0,0 +1,169 @@
...
+          },
+        },
+      },
+    },
+  } as unknown as state.State);
+
+const goOptions = {};
+
+describe('parsePriorityFee', () => {
+  test.each([
```
Siegrift: image

The title interpolation does not work for the second `test.each`. I recommend not using test.each but rather creating `it` with a dynamic title.

src/gas-prices.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    expect(throwingFunction).toThrow();
+  });
+});
+
+describe('getGasPrice', () => {
+  const baseFeePerGas = ethers.BigNumber.from('93000000000');
+  const maxPriorityFeePerGas = BigNumber.from(PRIORITY_FEE_IN_WEI);
+  const maxFeePerGas = baseFeePerGas.mul(BASE_FEE_MULTIPLIER).add(maxPriorityFeePerGas);
+  const testGasPrice = ethers.BigNumber.from('48000000000');
+  // Must be run also outside the `beforeEach` because Jest doesn't consider tests run with `test.each` as tests
```
Siegrift: This is not true, the `beforeEach` hook does run before the `test.each` and you should be using mocks and spies in beforeEach hook or directly in a test otherwise they won't work. See: https://github.com/facebook/jest/issues/9131

The reason why you need to have it defined there is because you call `createEip1559Providers` outside of a test (and before each hook), so without calling it there you create it using ethers and not the mocks which you want.

src/update-beacons.test.ts
```diff
@@ -0,0 +1,257 @@
...
+
+// Can't compare RPC provider instances so comparing groups where the provider is represented by its URL
+type ComparableProviderSponsorBeacons = {
+  provider: string;
+  sponsorAddress: string;
+  updateInterval: number;
+  beacons: BeaconUpdate[];
+};
+
+const cpsbg: ComparableProviderSponsorBeacons[] = [
```
Siegrift, line 155: ajldfasbkf

Siegrift (approved):
Found a few minor things, but otherwise 👍 LGTM.

Also, the PR title is inaccurate since this PR only contains unit tests.

amarthadan:
> Also, the PR title is inaccurate since this PR only contains unit tests.

What are you talking about? the last commit is adding an E2E test. Granted, just one, but still :smile:

Siegrift (commented):
🤦 
image

There was no scrollbar so I haven't reviewed the last commit.

test/fixtures/config.ts
```diff
@@ -55,42 +56,36 @@ export const buildAirseekerConfig = () => ({
   triggers: {
     beaconUpdates: {
       '31337': {
-        '0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC': {
+        '0x20557D8e841d7c2B12DBD45022B9147286F57691': {
           beacons: [
+            // {
```
Siegrift: Why is this commented? Can we remove it?

Siegrift:
@amarthadan can we finish this? I think this is in a good state for some time and just needs a few minutes of love/time :)

amarthadan:
> @amarthadan can we finish this? I think this is in a good state for some time and just needs a few minutes of love/time :)

I know but it's quite low on my list of priorities right now :sweat_smile: AFAIK the only thing that needs to be addressed is the `beforeEach`/`test.each` issue. Otherwise, it works, I've rebased and resolved conflicts recently. So, if you want you can finish it or we can merge it as it is right now.

Siegrift:
> I know but it's quite low on my list of priorities right now 😅 AFAIK the only thing that needs to be addressed is the beforeEach/test.each issue. Otherwise, it works, I've rebased and resolved conflicts recently. So, if you want you can finish it or we can merge it as it is right now.

I think it can wait a few more days but it should be done with that `test.each` change, I just want to avoid this hanging too long for no reason.

PR #38, opened 2022-04-25
Fix bad JS math

amarthadan:
When playing with Airseeker I run into this exceptionL
`[Error: underflow [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT-underflow ] (fault="underflow", operation="BigNumber.from", value=14000000.000000002, code=NUMERIC_FAULT, version=bignumber/5.6.0)]`
It's happening because JS math is horrible and doing for example `1e8 * 0.14` results in `14000000.000000002` which when passed to `ethers.BigNumber.from` causes an exception.


vponline (approved):
Nice :exploding_head:

acenolaza:
LOL 😁 https://github.com/api3dao/airseeker/pull/26#discussion_r849760362

PR #45, opened 2022-04-26
Unquoted docker secrets

Siegrift:
I tried to run Airseeker via docker with example the secrets file, but was getting an error during secrets interpolation. It felt strange since it worked fine when running it locally via `yarn start`.

This PR does:
- Adds a new example configuration for docker-secrets.env
- Adds a command in package.json to run the built Airseeker docker locally
- Improves error handling during secrets interpolation


amarthadan:
We may have some flaky tests in Airseeker as it runs fine locally and I don't really see how your changes would cause that.

amarthadan (approved):

Siegrift:
@amarthadan it might be because the PR is not up to date with master 🤷 ? I'll try to make them pass before merging.

PR #57, opened 2022-04-28
Use only unquoted secrets for Airseeker

amarthadan:
Having two different versions of the secrets file is quite annoying so I updated the `run-locally.sh` file to deal with unquoted secrets since that format is required by both Docker and our Terraform recipes.

Can somebody test it on Mac though? :smile: 

Tagging @aquarat and @Ashar2shahid to let you know that this is the format that will be needed for deployment.


config/secrets.example.env
```diff
@@ -1,10 +1,10 @@
-# Secrets passed to Airseeker run locally must be quoted.
+# Secrets passed to dockerized Airseeker must NOT be quoted.
```
Siegrift:
```suggestion
# Airseeker secrets must NOT be quoted.
```

scripts/run-locally.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/sh
 
+SECRETS_FILE=$(dirname $(readlink -f "$0"))/../config/secrets.env
```
Siegrift: I realized we want to add the fail immediatelly flag `set -e` (https://intoli.com/blog/exit-on-errors-in-bash-scripts/)

Siegrift (changes requested):
Unfortunately, does not work for me :/ (tried both bash and zsh).
When I do `yarn start`:
```
$ scripts/run-locally.sh ts-node src/index.ts
/Users/emanueltesar/vacuumlabs/api3/airseeker/scripts/../config/secrets.env
grep -v ^\s*# /Users/emanueltesar/vacuumlabs/api3/airseeker/scripts/../config/secrets.env
scripts/run-locally.sh: line 12: syntax error near unexpected token `('
error Command failed with exit code 2.
```

I tried to debug this a bit, but I am not sure why it happens. When I do:
```
cat <(grep -v "^\s*#" /Users/emanueltesar/vacuumlabs/api3/airseeker/scripts/../config/secrets.env | sed 's/\([^=]*\)=\(.*\)/\1=\"\2"/')
```

it works. Do we need this to be a shell script? I think it would be easier to make this a node script instead? Or use https://github.com/google/zx

You also want to change the `docker:run` command.

amarthadan:
@Siegrift I switched to `bash` can you try it now?

Siegrift:
Unfortunately it still doesn't work for me. 

When I do `yarn start` it starts the app, however the secrets are not passed correctly and I receive validation error `Error: Error interpolating secrets. Make sure the secrets format is correct. ReferenceError: AIRSEEKER_WALLET_MNEMONIC is not defined`. I've checked and `AIRSEEKER_WALLET_MNEMONIC` is defined in secrets, but undefined in Airseeker.

It seems the `source` command doesn't work properly, because the the secret is undefined even when I do `echo $AIRSEEKER_WALLET_MNEMONIC`

amarthadan:
Ok, this doesn't lead anywhere. @Siegrift can you take over this? It makes no sense for me to come up with a solution that might work on mac :smile: And I'm pretty sure that whatever will eventually work on mac will work on linux as well.

Siegrift:
Will do. Created: https://api3dao.atlassian.net/browse/BEC-398

Siegrift:
@amarthadan I've implemented it using JS as it was the fastest way to get this done. It seems to be working on my side.

package.json
```diff
@@ -35,12 +35,12 @@
     "eslint:fix": "eslint . --ext .js,.ts --fix",
     "postinstall": "husky install",
     "pm2:logs": "pm2 logs",
-    "pm2:start": "scripts/run-locally.sh pm2 start airseeker-service.config.js",
+    "pm2:start": "ts-node -T scripts/run-locally.ts pm2 start airseeker-service.config.js",
```
amarthadan, line 38: Why would you add the `-T` option? That's skipping the type checking, right?

scripts/run-locally.ts
```diff
@@ -0,0 +1,12 @@
+import { parse } from 'dotenv';
```
amarthadan, line 1: This is such a heavy-weight solution for such a tiny problem. Adding two more libraries, running node and TS and manually reading the secrets and adding the quotes :see_no_evil: I know none of that matters in this case but it's horrible :see_no_evil: My solution was so elegant. Stupid Mac... :sob:

package.json
Siegrift: Yes, but it's faster. I think that is more important. (But I can remove if you disagree)

scripts/run-locally.ts
Siegrift: I disliked your solution :p :D

package.json
amarthadan: Yeah, I'd say we want the type check pretty much all the time.

Siegrift (approved):
I've tested it both locally and using docker and got the same behaviour, and both seem to be working fine.

package.json
Siegrift: The scripts will be typechecked once https://github.com/api3dao/airseeker/pull/102 is merged.

PR #58, opened 2022-04-28
Add check that the signed data are not older than data on-chain

Siegrift:
https://api3dao.atlassian.net/browse/BEC-341

Please, start with reviewing `src/update-beacons.ts`. I've moved the on chain value fetching to the `updateBeacons` function and am passing it to the `checkSignedDataFreshness` (implemented now) and `checkUpdateCondition`.

The functions ensures the gateway timestamp is larger than on chain timestamp (both timestamps are UNIX timestamps).

### DIFF
--- a/src/check-condition.ts
+++ b/src/check-condition.ts
@@ -16,6 +16,11 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
   return absoluteDelta.mul(ethers.BigNumber.from(HUNDRED_PERCENT)).div(absoluteInitialValue);
 };
 
+export interface OnChainBeaconData {
+  value: ethers.BigNumber;
+  timestamp: number;
+}
+
 export const checkUpdateCondition = async (
   onChainData: OnChainBeaconData,
   deviationThreshold: number,
@@ -29,11 +34,6 @@ export const checkUpdateCondition = async (
 
   return updateInPercentage.gt(threshold);
 };
-
-export interface OnChainBeaconData {
-  value: ethers.BigNumber;
-  timestamp: number;
-}
 
 /**
  * Returns true when the signed data response is fresh enough to be used for an on chain update.
--- a/src/check-condition.ts
+++ b/src/check-condition.ts
@@ -1,5 +1,4 @@
 import { ethers } from 'ethers';
-import { SignedData } from './validation';
 
 // Number that represents 100% is chosen to avoid overflows in DapiServer's
 // `calculateUpdateInPercentage()`. Since the reported data needs to fit
@@ -16,18 +15,12 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
   return absoluteDelta.mul(ethers.BigNumber.from(HUNDRED_PERCENT)).div(absoluteInitialValue);
 };
 
-export interface OnChainBeaconData {
-  value: ethers.BigNumber;
-  timestamp: number;
-}
-
 export const checkUpdateCondition = async (
-  onChainData: OnChainBeaconData,
+  onChainVale: ethers.BigNumber,
   deviationThreshold: number,
   apiValue: ethers.BigNumber
 ): Promise<boolean> => {
-  const { value, timestamp: _timestamp } = onChainData;
-  const updateInPercentage = calculateUpdateInPercentage(value, apiValue);
+  const updateInPercentage = calculateUpdateInPercentage(onChainVale, apiValue);
   const threshold = ethers.BigNumber.from(Math.trunc(deviationThreshold * HUNDRED_PERCENT)).div(
     ethers.BigNumber.from(100)
   );
@@ -42,6 +35,6 @@ export const checkUpdateCondition = async (
  * https://github.com/api3dao/airnode-protocol-v1/blob/e0d778fabff0df888987a6db31498c93ee2f6219/contracts/dapis/DapiServer.sol#L867
  * This can happen if the gateway or Airseeker is down and Airkeeper does the updates instead.
  */
-export const checkSignedDataFreshness = (onChainData: OnChainBeaconData, signedData: SignedData) => {
-  return parseInt(signedData.data.timestamp, 10) > onChainData.timestamp;
+export const checkSignedDataFreshness = (onChainTimestamp: number, signedDataTimestamp: string) => {
+  return onChainTimestamp < parseInt(signedDataTimestamp, 10);
 };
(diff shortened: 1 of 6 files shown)


src/check-condition.ts
```diff
@@ -19,27 +17,31 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
...
+
+/**
+ * Returns true when the signed data response is fresh enough to be used for an on chain update.
+ *
+ * Update transaction with stale data would revert on chain, draining the sponsor wallet. See:
+ * https://github.com/api3dao/airnode-protocol-v1/blob/e0d778fabff0df888987a6db31498c93ee2f6219/contracts/dapis/DapiServer.sol#L867
+ * This can happen if the gateway or Airseeker is down and Airkeeper does the updates instead.
+ */
+export const checkSignedDataFreshness = (onChainData: OnChainBeaconData, signedData: SignedData) => {
+  return parseInt(signedData.data.timestamp, 10) > onChainData.timestamp;
```
vponline: Just curious what is the 10 radix doing here? If the timestamp is already `Math.floor`-ed in Airnode?

Mostly asking in case we need to add it somewhere else as well. :thinking:

vponline (approved):
Nice :+1:

src/check-condition.ts
Siegrift: I need `parseInt` to convert that to a number. In JS a radix should always (as a good practice) be used because radix 10 is not a default. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

src/update-beacons.ts
```diff
@@ -243,8 +258,22 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorBeaco
...
 
     logger.log(
-      `Beacon with ID ${beaconUpdateData.beaconId} sucessfully updated with value ${newBeaconValue}. Tx hash ${tx.data.hash}.`
+      `Beacon with ID ${beaconUpdateData.beaconId} successfully updated with value ${newBeaconValue}. Tx hash ${tx.data.hash}.`
     );
     nonce++;
   }
 };
+
+export const readOnChainBeaconData = async (
```
amarthadan, line 271: Can you follow the convention of other functions that the function is not returning the "go" result but an actual value or `null`?

src/check-condition.ts
```diff
@@ -19,27 +17,31 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
...
     ethers.BigNumber.from(100)
   );
 
   return updateInPercentage.gt(threshold);
 };
+
+export interface OnChainBeaconData {
+  value: ethers.BigNumber;
+  timestamp: number;
+}
```
amarthadan: Nit. Can you move the interface above the function? I think it's better for readability (or at least it helps me :sweat_smile:)

src/check-condition.ts
```diff
@@ -19,27 +17,31 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
 };
 
 export const checkUpdateCondition = async (
-  voidSigner: ethers.VoidSigner,
-  dapiServer: DapiServer,
-  beaconId: string,
+  onChainData: OnChainBeaconData,
```
amarthadan: Why do you pass the whole `OnChainBeaconData` object if we don't care about the timestamp? You can pass just the value itself.

src/check-condition.ts
```diff
@@ -19,27 +17,31 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
...
+}
+
+/**
+ * Returns true when the signed data response is fresh enough to be used for an on chain update.
+ *
+ * Update transaction with stale data would revert on chain, draining the sponsor wallet. See:
+ * https://github.com/api3dao/airnode-protocol-v1/blob/e0d778fabff0df888987a6db31498c93ee2f6219/contracts/dapis/DapiServer.sol#L867
+ * This can happen if the gateway or Airseeker is down and Airkeeper does the updates instead.
+ */
+export const checkSignedDataFreshness = (onChainData: OnChainBeaconData, signedData: SignedData) => {
```
amarthadan: Again, why pass the whole thing, we need just a timestamp.

amarthadan: And this one is especially confusing because you're passing two different structures but you're picking just timestamps from them.

amarthadan (changes requested):
I like the extraction of reading the on-chain value from `checkUpdateCondition` :+1:

src/check-condition.ts
vponline: We (will) need `timestamp` for #60 (but I guess it depends on who merges first :smile: )

amarthadan: I reacted in https://github.com/api3dao/airseeker/pull/60#discussion_r863726412

src/update-beacons.ts
Siegrift: Sure.

src/check-condition.ts
Siegrift: https://www.youtube.com/watch?v=Y0pdQU87dc8

src/check-condition.ts
Siegrift: It is not that confusing to me. The `updateBeacons` functions is complex and it's doing a lot more then it should. I am delegating that part of the logic to this helper function.

Siegrift: But It is now inconsistent with the other functions so I'll update it.

src/check-condition.ts
```diff
@@ -19,27 +16,25 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
...
+/**
+ * Returns true when the signed data response is fresh enough to be used for an on chain update.
+ *
+ * Update transaction with stale data would revert on chain, draining the sponsor wallet. See:
+ * https://github.com/api3dao/airnode-protocol-v1/blob/e0d778fabff0df888987a6db31498c93ee2f6219/contracts/dapis/DapiServer.sol#L867
+ * This can happen if the gateway or Airseeker is down and Airkeeper does the updates instead.
+ */
+export const checkSignedDataFreshness = (onChainTimestamp: number, signedDataTimestamp: string) => {
+  return onChainTimestamp < parseInt(signedDataTimestamp, 10);
+};
```
amarthadan, line 40: I'm wondering if this function even makes sense since it's just one comparison. Maybe we should just do that directly in `updateBeacons`. But that's just a nit, I'm ok with having it here.

amarthadan (approved):

src/check-condition.ts
Siegrift: It is easier to test this way. It's better to keep it imo.

PR #59, opened 2022-04-29
Configure release

aquarat:
This configures release v0.2.0.


amarthadan (approved):

package.json
```diff
@@ -1,9 +1,18 @@
 {
-  "name": "airseeker",
+  "name": "@api3/airseeker",
   "license": "MIT",
-  "version": "1.0.0",
-  "description": "A tool to update a beacons with signed responses from Airnode's gateway",
+  "version": "0.2.0",
+  "description": "A tool to update beacons with signed responses from remote Airnode gateways",
   "main": "./dist/src/main.js",
+  "types": "./dist/src/validation",
```
Siegrift, line 7: We do not need `types` since no one will use Airseeker as a library.

amarthadan: I think @aquarat added that on purpose as he plans to use the types in the `ops` repo.

aquarat: We do validation in ops - so the types will be useful for that :)

PR #60, opened 2022-05-02
Add heartbeatInterval check to checkCondition

vponline:
Update the beacon if the Dapi Server timestamp value is older than the `heartbeatInterval` setting.

### DIFF
--- a/src/check-condition.test.ts
+++ b/src/check-condition.test.ts
@@ -2,6 +2,7 @@ import { BigNumber, ethers } from 'ethers';
 import {
   calculateUpdateInPercentage,
   checkSignedDataFreshness,
+  checkOnchainDataFreshness,
   checkUpdateCondition,
   HUNDRED_PERCENT,
   OnChainBeaconData,
@@ -89,7 +90,7 @@ describe('checkUpdateCondition', () => {
 });
 
 describe('checkSignedDataFreshness', () => {
-  it('returns true if signed data gateway is newer then on chain record', () => {
+  it('returns true if signed data gateway is newer than on chain record', () => {
     const isFresh = checkSignedDataFreshness(
       { value: BigNumber.from('123'), timestamp: getUnixTimestamp('2022-4-28') },
       validSignedData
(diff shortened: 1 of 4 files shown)


acenolaza (approved):
LGTM 👍🏻

src/check-condition.ts
```diff
@@ -35,7 +36,10 @@ export const checkUpdateCondition = async (
     return null;
   }
 
-  const [dapiServerValue, _timestamp] = goDataFeed.data;
+  const [dapiServerValue, dapiServerTimestamp] = goDataFeed.data;
+
+  if (Date.now() / 1000 - heartbeatInterval > dapiServerTimestamp) return true;
+
```
amarthadan: This was much easier than I anticipated :joy: 

But as a reaction to https://github.com/api3dao/airseeker/pull/58#discussion_r863719331 I was about to ask you if you can rebase this on top of the `signed-data-freshness` branch and do this check outside `checkUpdateCondition`.
Why I want this is basically distinguishing when we're updating because of the heartbeat and when we're updating because of the deviation threshold. And log these two reasons accordingly. WDYT?

vponline: Makes perfect sense , I'll do that

src/check-condition.test.ts
```diff
@@ -89,7 +90,7 @@ describe('checkUpdateCondition', () => {
 });
 
 describe('checkSignedDataFreshness', () => {
-  it('returns true if signed data gateway is newer then on chain record', () => {
+  it('returns true if signed data gateway is newer than on chain record', () => {
```
Siegrift, line 89: :D never trust my english.

Siegrift (approved):
👍 LGTM. 

I dislike the `updateBeacons` function as it is too complex and we should think about how to split it in the future.

amarthadan:
> I dislike the `updateBeacons` function as it is too complex and we should think about how to split it in the future.

I agree with you but when I tried to divide it (moving the processing of a single beacon into a separate function) it became even uglier so I left it like it is for now. Maybe we should revisit that.

src/update-beacons.ts
```diff
@@ -204,27 +204,41 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorBeaco
...
+      }
+      if (shouldUpdate === null) {
+        logger.log(`Unable to fetch current beacon value for beacon with ID ${beaconUpdateData.beaconId}.`);
+        // This can happen only if we reach the total timeout so it makes no sense to continue with the rest of the beacons
+        return;
+      }
+      if (!shouldUpdate) {
+        logger.log(`Deviation threshold not reached for beacon with ID ${beaconUpdateData.beaconId}. Skipping.`);
+        continue;
+      }
```
amarthadan: We don't need to do both `if(shouldUpdate)` and `if(!shouldUpdate)`.
```suggestion
      if (shouldUpdate === null) {
        logger.log(`Unable to fetch current beacon value for beacon with ID ${beaconUpdateData.beaconId}.`);
        // This can happen only if we reach the total timeout so it makes no sense to continue with the rest of the beacons
        return;
      }
      if (!shouldUpdate) {
        logger.log(`Deviation threshold not reached for beacon with ID ${beaconUpdateData.beaconId}. Skipping.`);
        continue;
      }
      
      logger.log(`Deviation threshold reached for beacon with ID ${beaconUpdateData.beaconId}. Updating.`);
```

amarthadan (approved):
:hammer_and_wrench: & :ship:

PR #73, opened 2022-05-02
BEC-363: Implement configurable logging

acenolaza:
Re-used the Airnode logger structure of using `meta` and `additional`.

### DIFF
--- a/src/fetch-beacon-data.ts
+++ b/src/fetch-beacon-data.ts
@@ -1,6 +1,10 @@
-import { go } from '@api3/promise-utils';
-import { ethers } from 'ethers';
 import { isEmpty, uniq } from 'lodash';
+import { ethers } from 'ethers';
+import { go } from '@api3/promise-utils';
+import { logger } from './logging';
+import { getState, updateState } from './state';
+import { makeSignedDataGatewayRequests } from './make-request';
+import { shortenAddress, sleep } from './utils';
 import {
   GATEWAY_TIMEOUT_MS,
   INFINITE_RETRIES,
@@ -8,10 +12,6 @@ import {
   RANDOM_BACKOFF_MAX_MS,
   RANDOM_BACKOFF_MIN_MS,
 } from './constants';
-import { logger } from './logging';
-import { makeSignedDataGatewayRequests } from './make-request';
-import { getState, updateState } from './state';
-import { shortenAddress, sleep } from './utils';
 
 export const initiateFetchingBeaconData = async () => {
   logger.debug('Initiating fetching all beacon data');
--- a/src/fetch-beacon-data.ts
+++ b/src/fetch-beacon-data.ts
@@ -1,10 +1,6 @@
-import { isEmpty, uniq } from 'lodash';
-import { ethers } from 'ethers';
 import { go } from '@api3/promise-utils';
-import { logger } from './logging';
-import { getState, updateState } from './state';
-import { makeSignedDataGatewayRequests } from './make-request';
-import { sleep } from './utils';
+import { ethers } from 'ethers';
+import { isEmpty, uniq } from 'lodash';
 import {
   GATEWAY_TIMEOUT_MS,
   INFINITE_RETRIES,
@@ -12,20 +8,27 @@ import {
   RANDOM_BACKOFF_MAX_MS,
   RANDOM_BACKOFF_MIN_MS,
 } from './constants';
+import { logger } from './logging';
+import { makeSignedDataGatewayRequests } from './make-request';
+import { getState, updateState } from './state';
+import { shortenAddress, sleep } from './utils';
 
 export const initiateFetchingBeaconData = async () => {
-  logger.log('Initiating fetching all beacon data');
+  logger.debug('Initiating fetching all beacon data');
   const { config } = getState();
 
   const beaconIdsToUpdate = uniq(
-    Object.values(config.triggers.beaconUpdates).flatMap((beaconUpdatesPerSponsor) => {
-      return Object.values(beaconUpdatesPerSponsor).flatMap((beaconUpdate) => {
+    Object.entries(config.triggers.beaconUpdates).flatMap(([chainId, beaconUpdatesPerSponsor]) => {
+      return Object.entries(beaconUpdatesPerSponsor).flatMap(([sponsorAddress, beaconUpdate]) => {
         const { beacons } = beaconUpdate;
         // TODO: Should be later part of the validation
         const foundBeacons = beacons.filter((beacon) => {
           if (config.beacons[beacon.beaconId]) return true;
 
-          logger.log(`Missing beacon with ID ${beacon.beaconId}. Skipping.`);
+          logger.warn(`Missing beacon with ID ${beacon.beaconId}. Skipping.`, {
+            meta: { chainId },
+            additional: { Sponsor: shortenAddress(sponsorAddress) },
+          });
           return false;
         });
         return foundBeacons.flatMap((b) => b.beaconId);
@@ -34,7 +37,7 @@ export const initiateFetchingBeaconData = async () => {
   );
 
   if (isEmpty(beaconIdsToUpdate)) {
-    logger.log('No beacons to fetch data for found. Stopping.');
+    logger.error('No beacons to fetch data for found. Stopping.');
     process.exit(NO_FETCH_EXIT_CODE);
   }
 
@@ -67,7 +70,8 @@ export const fetchBeaconDataInLoop = async (beaconId: string) => {
 };
 
 export const fetchBeaconData = async (beaconId: string) => {
-  logger.log(`Fetching beacon data for: ${beaconId}`);
+  const logOptionsBeaconId = { additional: { 'Beacon-ID': beaconId } };
+  logger.debug('Fetching beacon data', logOptionsBeaconId);
   const { config } = getState();
 
   const { fetchInterval, airnode, templateId } = config.beacons[beaconId];
@@ -80,19 +84,20 @@ export const fetchBeaconData = async (beaconId: string) => {
     [template.endpointId, template.parameters]
   );
   if (derivedTemplateId !== templateId) {
-    logger.log(`Invalid template ID ${templateId}. Skipping.`);
+    logger.warn(`Invalid template ID ${templateId}. Skipping.`, logOptionsBeaconId);
     return;
   }
 
-  const goRes = await go(() => makeSignedDataGatewayRequests(gateway, template), {
+  const goRes = await go(() => makeSignedDataGatewayRequests(gateway, { ...template, id: templateId }), {
     attemptTimeoutMs: GATEWAY_TIMEOUT_MS,
     retries: INFINITE_RETRIES,
     delay: { type: 'random', minDelayMs: RANDOM_BACKOFF_MIN_MS, maxDelayMs: RANDOM_BACKOFF_MAX_MS },
     totalTimeoutMs: fetchInterval * 1_000,
-    onAttemptError: (goError) => logger.log(`Failed attempt to call signed data gateway. Error: ${goError.error}`),
+    onAttemptError: (goError) =>
+      logger.warn(`Failed attempt to call signed data gateway. Error: ${goError.error}`, logOptionsBeaconId),
   });
   if (!goRes.success) {
-    logger.log(`Unable to call signed data gateway. Error: "${goRes.error}"`);
+    logger.warn(`Unable to call signed data gateway. Error: "${goRes.error}"`, logOptionsBeaconId);
     return;
   }
 
--- a/src/logging.ts
+++ b/src/logging.ts
@@ -1,11 +1,19 @@
-import { logger as airnodeLogger } from '@api3/airnode-utilities';
+import { logger as airnodeLogger, LogOptions } from '@api3/airnode-utilities';
+import merge from 'lodash/merge';
 import { getState } from './state';
 
-const debug = (message: string) => airnodeLogger.debug(message, getState().logOptions);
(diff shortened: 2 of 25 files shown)


src/check-condition.ts
```diff
@@ -26,12 +26,15 @@ export const checkUpdateCondition = async (
   apiValue: ethers.BigNumber,
   goOptions: GoAsyncOptions
 ): Promise<boolean | null> => {
+  const logOptionsBeaconId = { additional: { 'Beacon-ID': beaconId, 'Dapi-Server': dapiServer.address } };
```
amarthadan: I think we would need to add chain ID as well because you can have the same beacon ID and dapi server address on multiple chains (I think). You don't need to add it here (as we don't have it available) but I think https://github.com/api3dao/airseeker/pull/58 changes the structure a bit and it will become easy to add. Just add a TODO for now.

src/fetch-beacon-data.ts
```diff
@@ -1,31 +1,34 @@
...
   GATEWAY_TIMEOUT_MS,
   INFINITE_RETRIES,
   NO_FETCH_EXIT_CODE,
   RANDOM_BACKOFF_MAX_MS,
   RANDOM_BACKOFF_MIN_MS,
 } from './constants';
+import { logger } from './logging';
+import { makeSignedDataGatewayRequests } from './make-request';
+import { getState, updateState } from './state';
+import { shortenAddress, sleep } from './utils';
```
amarthadan: Why did the order change? If your editor is doing this automatically please, turn it off because it's creating these unnecessary changes.

src/fetch-beacon-data.ts
```diff
@@ -1,31 +1,34 @@
...
+      return Object.entries(beaconUpdatesPerSponsor).flatMap(([sponsorAddress, beaconUpdate]) => {
         const { beacons } = beaconUpdate;
         // TODO: Should be later part of the validation
         const foundBeacons = beacons.filter((beacon) => {
           if (config.beacons[beacon.beaconId]) return true;
 
-          logger.log(`Missing beacon with ID ${beacon.beaconId}. Skipping.`);
+          logger.warn(`Missing beacon with ID ${beacon.beaconId}. Skipping.`, {
+            meta: { chainId },
+            additional: { Sponsor: shortenAddress(sponsorAddress) },
```
amarthadan, line 30: What's the difference between `meta` and `additional`?

src/logging.ts
```diff
@@ -1,11 +1,19 @@
...
 
-const debug = (message: string) => airnodeLogger.debug(message, getState().logOptions);
-const error = (message: string) => airnodeLogger.error(message, getState().logOptions);
-const info = (message: string) => airnodeLogger.info(message, getState().logOptions);
-const log = (message: string) => airnodeLogger.log(message, getState().logOptions);
-const warn = (message: string) => airnodeLogger.warn(message, getState().logOptions);
+type LogOptionsOverride = Partial<Pick<LogOptions, 'meta' | 'additional'>>;
+
+const debug = (message: string, logOptionsOverride?: LogOptionsOverride) =>
+  airnodeLogger.debug(message, merge({ ...getState().logOptions }, logOptionsOverride));
```
amarthadan, line 8: Do we actually need the `merge` function? If I understand it correctly, `LogOptionsOverride` can have only `meta` and `additional` properties. We don't have these in the state log options. So, simple
```ts
{...getState().logOptions, ...logOptionsOverride}
```
should be enough.

src/state.ts
```diff
@@ -28,7 +32,10 @@ export const initializeState = (config: Config) => {
     stopSignalReceived: false,
     beaconValues: {},
     providers: {},
-    logOptions: DEFAULT_LOG_OPTIONS,
+    logOptions: buildBaseOptions(
+      { nodeSettings: { logFormat: config.logFormat, logLevel: config.logLevel } },
+      { coordinatorId: randomHexString(16) }
```
amarthadan, line 37: I'm fine with this even though this will change only when the Airseeker is restarted which shouldn't happen very often. But I would name it differently, what about `airseekerId`?

src/update-beacons.ts
```diff
@@ -203,12 +211,12 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorBeaco
...
     );
     if (shouldUpdate === null) {
-      logger.log(`Unable to fetch current beacon value for beacon with ID ${beaconUpdateData.beaconId}.`);
+      logger.warn(`Unable to fetch current beacon value`, logOptionsBeaconId);
       // This can happen only if we reach the total timeout so it makes no sense to continue with the rest of the beacons
       return;
     }
     if (!shouldUpdate) {
-      logger.log(`Deviation threshold not reached for beacon with ID ${beaconUpdateData.beaconId}. Skipping.`);
+      logger.log(`Deviation threshold not reached. Skipping.`, logOptionsBeaconId);
```
amarthadan: What level is `log()` logging under?

src/update-beacons.ts
```diff
@@ -231,19 +239,19 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorBeaco
...
-      logger.log(
-        `Unable to update beacon with ID ${beaconUpdateData.beaconId} using wallet ${sponsorWallet.address} and nonce ${nonce}. Error: ${tx.error}`
-      );
+      logger.warn(`Unable to update beacon with nonce ${nonce}. Error: ${tx.error}`, logOptionsBeaconId);
       return;
     }
 
-    logger.log(
-      `Beacon with ID ${beaconUpdateData.beaconId} sucessfully updated with value ${newBeaconValue}. Tx hash ${tx.data.hash}.`
+    logger.warn(
```
amarthadan: This is definitely not a warning :sweat_smile:

src/validation.ts
```diff
@@ -77,6 +80,8 @@ export const triggersSchema = z.object({
 export const configSchema = z
   .object({
     airseekerWalletMnemonic: z.string(),
+    logFormat: logFormatSchema,
+    logLevel: logLevelSchema,
```
amarthadan: Maybe it would be better to have one `logging` or `log` section and having `format` and `level` there:
```json
{
  "log": {
    "format": "plain",
    "level": "INFO"
  }
}
```

amarthadan (changes requested):
I have two concerns:
* We're a bit inconsistent in what's considered a warning and what error. Are we considering warning to be everything that doesn't stop the Airseeker? Because due to the design that's almost everything. I don't mind we should just clarify the intentions.
* I think we shouldn't use `log()` when we're using `warn()`, `debug()`, etc. We should instead use `info()` to be explicit.

acenolaza:
> I have two concerns:
> 
> * We're a bit inconsistent in what's considered a warning and what error. Are we considering warning to be everything that doesn't stop the Airseeker? Because due to the design that's almost everything. I don't mind we should just clarify the intentions.

That is exactly right. At least that was how I approach it in Airkeeper but I'm open to comments and suggestions. That is also why there are just a few `logger.error` in the code where we are just throwing and stopping the Airseeker execution.

> * I think we shouldn't use `log()` when we're using `warn()`, `debug()`, etc. We should instead use `info()` to be explicit.

Completely agree. That was my mistake.

src/update-beacons.ts
acenolaza: 🤦🏻‍♂️ 😁  good catch!

src/update-beacons.ts
acenolaza: I should have switched it to `info()`

src/state.ts
acenolaza: Right, I faced this same limitation in Airkeeper. The problem is that [buildBaseOptions](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/logging/logging.ts#L11-L17) is a function imported from the airnode-utilities package and it creates a `LogOptions` object which only has `coordinatorId`. We might want to make it more dynamic on the [airnode-utilities type](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/logging/types.ts#L6). Something like `readonly [key:string]:string` or maybe just `readonly processId:string`

src/logging.ts
acenolaza: I started by doing that but then I realized I didn't want to override the fields currently in `meta` and `additional` child objects so that is why I decided to do a deep merge and only add instead of replace.

src/fetch-beacon-data.ts
acenolaza: [additional](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/logging/types.ts#L14) seems to be more flexible since it is of type `any`

src/fetch-beacon-data.ts
acenolaza: I'm doing that manually by clicking shift+alt+o to remove any unused import that I might have added while writing the code. I'll stop doing that if it's an issue.

src/check-condition.ts
acenolaza: I'll add the TODO then

src/fetch-beacon-data.ts
amarthadan: I see :confused: I think we will eventually change that in Airnode as it's currently built mainly for Airnode and doesn't really fit the other tools we're building. I would actually ignore the `meta` object entirely and add everything to the `additional`. That way, once we generalize that on Airnode (which I think will lead to keeping just one field) we'll easily switch. WDYT? I presume the way it's logged, the actual output, is the same for `meta` and `additional` right?

src/logging.ts
amarthadan: But that's what I'm saying. The logging options in the state don't have `meta` and `additional` fields so it should be safe to concat them just by spreading.

amarthadan:
> That is exactly right. At least that was how I approach it in Airkeeper but I'm open to comments and suggestions. That is also why there are just a few `logger.error` in the code where we are just throwing and stopping the Airseeker execution.

I'm fine with this approach, just wanted to make sure :slightly_smiling_face:

src/make-request.ts
```diff
@@ -38,14 +45,14 @@ export const makeSignedDataGatewayRequests = async (gateways: Gateway[], templat
 
     if (!goRes.success) {
       const message = `Failed to make signed data gateway request for gateway: "${fullUrl}". Error: "${goRes.error}"`;
-      logger.log(message);
+      logger.error(message, logOptionsTemplateId);
       throw new Error(message);
     }
```
amarthadan: Even though this one, and the other two below, are throwing errors, if you look at [`fetchBeaconData`](https://github.com/api3dao/airseeker/blob/main/src/fetch-beacon-data.ts#L87) you'll see that they are handled and Airseeker is not stopping. So, I think, those should be just warnings.

amarthadan (changes requested):

src/check-condition.ts
acenolaza: Since #58 has been merged I'll look into this now and hopefully remove the TODO

src/fetch-beacon-data.ts
acenolaza: ~Right, the actual output doesn't care if a value is inside `meta` or inside `additional` but WDYT if I just add `airseekerId` to additional and keep using `chainId` and `providerName` from meta (since those are probably not going to change)?~

src/logging.ts
acenolaza: Currently I'm setting the `coordinatorId` in `state.meta` when calling the [airnode-utitilities.buildBaseOptions](https://github.com/api3dao/airseeker/blob/db2518caa8161874afa15708ceed56e507a87d82/src/state.ts#L37) function that initializes the `meta` object.

I know that we will probably stop using the `meta.coordinatorId` field in favor of adding a new `additional.airseekerId` field but this field will probably be initialized when state is created, right? I mean we would still need to merge in order to avoid overriding `additional` object and loose the `airseekerId` value.

src/fetch-beacon-data.ts
acenolaza: Actually, I just found that airnode-utilities logger expects to have something in `meta` object when setting `additional`, otherwise get a comma at the beginning like this: 
image
because of [this](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/logging/logging.ts#L114) line in the `reduce` function.

One other thing is that we may want to tweak is that currently `meta` arg in `buildBaseOptions` [function](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/logging/logging.ts#L11) is not optional.

My suggestion is to keep using `meta.coordinatorId` until we make changes to the logging.ts file to support apps other than Airnode.

amarthadan: I agree, let's do that.

src/logging.ts
amarthadan: You're right, I'm just blind, carry on :sweat_smile:

src/update-beacons.ts
```diff
@@ -66,11 +66,11 @@ export const groupBeaconsByProviderSponsor = () => {
 };
 
 export const initiateBeaconUpdates = () => {
-  logger.log('Initiating beacon updates');
+  logger.debug('Initiating beacon updates');
 
   const providerSponsorBeaconsGroups = groupBeaconsByProviderSponsor();
   if (isEmpty(providerSponsorBeaconsGroups)) {
-    logger.log('No beacons for processing found. Stopping.');
+    logger.info('No beacons for processing found. Stopping.');
```
amarthadan: This should be logged as an error.

amarthadan (approved):
I think we don't need to do anything about the `meta` vs `additional` issue at the moment. I'm fine with the current state. Fix that one info/error issue and feel free to merge.

PR #74, opened 2022-05-04
Update Terraform recipes to reuse default VPC

amarthadan:
Other than reusing the default VPC and NAT gateway I did a bit of cleaning because we were adding some networking pieces that were not really needed.


aquarat (approved):
LGTM as per Slack and the functional Airseeker deployment. Thanks :)

PR #75, opened 2022-05-05
Add gas oracle for providers

vponline:
The gas oracle will fetch the latest and latest - 20 blocks and get the specified `percentile` from each block. The latest block's `percentile` will be used if the difference between the two percentiles does not exceed the `gasDeviationThreshold`.

`getGasPrice` will be used as a fallback if the `gasDeviationThreshold` is exceeded, either block has less than `minBlockTransactions` or failing to fetch either block.

The `backupGasPriceGwei` config option will be used if failing to fetch both blocks and the fallback `gasPrice`.


config/airseeker.example.json
```diff
@@ -28,6 +28,11 @@
           "unit": "gwei"
         },
         "baseFeeMultiplier": 2
+      },
+      "gasOracle": {
+        "sampleBlockCount": 20,
+        "percentile": 60,
+        "updateInterval": 20
       }
```
vponline: I put the config at the chain level, but we could do per provider configuration as well if needed. WDYT?

src/gas-oracle.ts
```diff
@@ -0,0 +1,195 @@
...
+    },
+  }));
+};
+
+const getPercentile = (percentile: number, array: ethers.BigNumber[]) => {
+  array.sort((a, b) => (a.gt(b) ? 1 : -1));
+
+  const index = Math.ceil(array.length * (percentile / 100)) - 1;
+  return array[index];
+};
```
vponline: @bbenligiray  Do we want something fancier here instead of just returning the `gasPrice` at the Pth percentile?

src/gas-oracle.ts
```diff
@@ -0,0 +1,195 @@
...
+      ? latestBlock.number - latestBlockNumberInState
+      : sampleBlockCount;
+
+  const newBlockData: BlockData[] = [];
+  // Add the latest block if it has transactions and skip otherwise
+  if (latestBlock.transactions.length) {
+    newBlockData.push({
+      blockNumber: latestBlock.number,
+      gasPrices: latestBlock.transactions.map(
+        (tx) => (tx.gasPrice || latestBlock.baseFeePerGas?.add(tx.maxPriorityFeePerGas!))!
```
vponline: Based on the discussion with @aquarat in slack, even EIP1559 chains seem to have the `gasPrice` field but I added `baseFeePerGas + maxPriorityFeePerGas` as backup.

bbenligiray:
As a note, the transaction contents of block #1234 may have changed since `updateInterval` ago because of something like a reorg. Therefore, this approach uses an approximate gas price history. This is fine for our current purposes, but we should be aware in case we want to use the same history for another purpose.

src/gas-oracle.ts
```diff
@@ -0,0 +1,195 @@
...
+  return array[index];
+};
+
+const fetchUpdateBlockData = async (provider: Provider, sampleBlockCount: number, percentile: number) => {
+  const { rpcProvider, chainId, providerName } = provider;
+  logger.log(`Fetching blocks on chain with ID ${chainId} for provider with name ${providerName}`);
+
+  // Get latest block
+  const goRes = await go(() => rpcProvider.getBlockWithTransactions('latest'), {
+    onAttemptError: (goError) => logger.log(`Failed attempt to get block. Error: ${goError.error}`),
```
vponline: I still need to add retry options. Should we go with the same settings as the current `getGasPrice` (i.e. `prepareGoOptions(startTime, totalTimeout)`)?

vponline: I added these and moved `prepareGoOptions` to utils

src/constants.ts
```diff
@@ -15,3 +15,9 @@ export const INT224_MIN = ethers.BigNumber.from(2).pow(ethers.BigNumber.from(223
 export const INT224_MAX = ethers.BigNumber.from(2).pow(ethers.BigNumber.from(223)).sub(ethers.BigNumber.from(1));
 export const NO_BEACONS_EXIT_CODE = 1;
 export const NO_FETCH_EXIT_CODE = 2;
+export const NO_ORACLE_EXIT_CODE = 3;
+export const DEFAULT_GAS_ORACLE_UPDATE_INTERVAL = 20;
+export const DEFAULT_GAS_PRICE_PERCENTILE = 60;
+export const DEFAULT_SAMPLE_BLOCK_COUNT = 20;
+// The gas price to use if the gas oracle fails to fetch any values from the provider and no value is specified in airseeker.json
+export const DEFAULT_BACK_UP_GAS_PRICE_GWEI = 10;
```
acenolaza: I might be wrong but adding the `DEFAULT_` prefix might be not consistent with the other consts which are also used as default config values.

src/update-beacons.test.ts
```diff
@@ -247,7 +250,7 @@ describe('calculateTimeout', () => {
     const totalTimeout = 3000;
     jest.spyOn(Date, 'now').mockReturnValue(1650548023000);
 
-    expect(api.calculateTimeout(startTime, totalTimeout)).toEqual(2000);
```
acenolaza, line 250: This test and the one in line 254 seem like they need to be moved to a utils.test.ts file

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+    await fetchUpdateBlockData(provider, updateInterval, sampleBlockCount, percentile, backupGasPriceGwei);
+
+    const duration = Date.now() - startTimestamp;
+    const waitTime = Math.max(0, updateInterval * 1_000 - duration);
+    await sleep(waitTime);
+  }
+};
+
+export const initiateFetchingBlockData = () => {
+  logger.log(`Initiating gas oracles`);
```
acenolaza: We have stopped using `logger.log` and you should now start sending a logOptions object as a second arg to logger functions.

You can check gas-prices.ts or any other file to use as reference.

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+};
+
+export const fetchBlockDataInLoop = async (provider: Provider) => {
+  const { config } = getState();
+  const gasOracleConfig = config.chains[provider.chainId].gasOracle;
+
+  const updateInterval = gasOracleConfig?.updateInterval || DEFAULT_GAS_ORACLE_UPDATE_INTERVAL;
+  const sampleBlockCount = gasOracleConfig?.sampleBlockCount || DEFAULT_SAMPLE_BLOCK_COUNT;
+  const percentile = gasOracleConfig?.percentile || DEFAULT_GAS_PRICE_PERCENTILE;
+  const backupGasPriceGwei = gasOracleConfig?.backupGasPriceGwei || DEFAULT_BACK_UP_GAS_PRICE_GWEI;
```
acenolaza: A small improvement could be to do this only once per chain instead of per provider since the config is set per chain.

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+
+export const fetchUpdateBlockData = async (
+  provider: Provider,
+  updateInterval: number,
+  sampleBlockCount: number,
+  percentile: number,
+  backupGasPriceGwei: number
+) => {
+  const { rpcProvider, chainId, providerName } = provider;
+  logger.log(`Fetching blocks on chain with ID ${chainId} for provider with name ${providerName}`);
```
acenolaza: You may want to reword these error messages once `chainId` and `providerName` are added to the `logOptions` object

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+        !state.gasOracles[chainId][providerName].backupGasPrice
+      ) {
+        updateBackupGasPrice(chainId, providerName, ethers.utils.parseUnits(backupGasPriceGwei.toString(), 'gwei'));
+      }
+      return;
+    }
+
+    // Add back up gasTarget to state
+    const { gasPrice, maxFeePerGas } = feeDataRes.data;
+    updateBackupGasPrice(chainId, providerName, gasPrice || maxFeePerGas);
```
acenolaza: Could ethers ever return both `gasPrice` and `maxFeePerGas`? I ask because we might need to read the `chains.options.txType` config option to determine which value to use.

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+  if (latestBlockInStateMatch) {
+    logger.log(
+      `Latest block on chain with ID ${chainId} for provider with name ${providerName} already in state. Skipping.`
+    );
+    return;
+  }
+
+  // Calculate how many blocks to fetch with a maximum of sampleBlockCount
+  const blockCountToFetch =
+    latestBlock.number - latestBlockNumberInState <= sampleBlockCount
```
acenolaza: Is it possible that the `latestBlock.number` might be the same as the `latestBlockNumberInState` 🤔 If so then `blockCountToFetch` might be 0

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+      ? latestBlock.number - latestBlockNumberInState
+      : sampleBlockCount;
+
+  const newBlockData: BlockData[] = [];
+  // Add the latest block if it has transactions and skip otherwise
+  if (latestBlock.transactions.length) {
+    newBlockData.push({
+      blockNumber: latestBlock.number,
+      gasPrices: latestBlock.transactions.map(
+        (tx) => (tx.gasPrice || latestBlock.baseFeePerGas?.add(tx.maxPriorityFeePerGas!))!
```
acenolaza: I guess here we should use ! instead of ?, right?

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+      ),
+    });
+  }
+
+  // Fetch additional blocks up to a limit of blockCountToFetch
+  let loopIndex = 1;
+  while (newBlockData.length < blockCountToFetch) {
```
acenolaza: Should we just use <= to get rid of lines 183-190?

acenolaza: Nit: we could maybe also use a for loop in order to avoid creating a `loopIndex` var

src/gas-oracle.ts
```diff
@@ -0,0 +1,260 @@
...
+        [providerName]: { blockData, percentileGasPrice },
+      },
+    },
+  }));
+};
+
+export const getPercentile = (percentile: number, array: ethers.BigNumber[]) => {
+  array.sort((a, b) => (a.gt(b) ? 1 : -1));
+
+  const index = Math.ceil(array.length * (percentile / 100)) - 1;
```
acenolaza, line 34: I wonder if there's some library that we can use here 🤔 Probably not one that supports `ethers.BigNumber[]` 😁 

A few things that worry me are:
1. can percentile be 0?
2. can array be empty?

acenolaza (changes requested):
The code comments are really helpful for reading the code 👏🏻 You may want to consider adding a description about the new config options in the README.

config/airseeker.example.json
bbenligiray: Chain level imo, same as the old gas price parameters

src/gas-oracle.ts
bbenligiray: No, this is fine. Array can be empty though? (I see that Santiago pointed that out too.)

bbenligiray:
This is simpler to implement, but is it possible to not have an `updateInterval` parameter and only update the gas price oracle if there is a transaction to be made?

vponline:
> This is simpler to implement, but is it possible to not have an `updateInterval` parameter and only update the gas price oracle if there is a transaction to be made?

I don't see why this wouldn't also work. Should I make the changes?

src/gas-oracle.ts
vponline: > Probably not one that supports ethers.BigNumber

Yep, otherwise there are quite a few library options. I took the implenentation from here: [percentile](https://github.com/d4rkr00t/percentile/blob/master/scripts/percentile-baseline.js). (Although without all the number checks that it has :sweat_smile:)

> A few things that worry me are:
> 1. can percentile be 0?
> 2. can array be empty?

Good points! 
1. The validator checks that the percentile setting is not 0: `percentile: z.number().int().positive().optional()`
2. In practice, the code shouldn't go this far if it fails to get `blockData` (and go into the fallback gasPrice if-block) so it should never have an empty array, but you're right in that I should address these cases.

src/gas-oracle.ts
vponline: > Should we just use <= to get rid of lines 183-190?

Hmm I may not understand what you mean, but wouldn't this re-fetch the 'latest' block?

src/gas-oracle.ts
vponline: Yep and then it shouldn't proceed to fetch more blocks, although it _should_ return from the function in this case from line 168:

`if (latestBlockInStateMatch) {...}`

But do you see a potential issue with this logic? :thinking:

bbenligiray:
> I don't see why this wouldn't also work. Should I make the changes?

I think so. Keeping a gas price oracle ready to go in case we need to make a transaction is not useful because it goes stale so fast. For the same reason, we don't really want to use a 15 second-old gas price esitmation.

src/gas-oracle.ts
vponline: I tried this on a few chains and it actually always returns both (even if it's an EIP1559 network). It seems that `gasPrice` is always returned and could only be `null` if there is an error.

My understading was that we want to always use the (legacy) gasPrice value so I only added the `|| maxFeePerGas` in case ethers has an error getting the gas price. Please let me know if you had another thought though. 

Here's the `ethers.getFeeData()` function for reference of how they implement it:

```ts
async getFeeData(): Promise<FeeData> {
        const { block, gasPrice } = await resolveProperties({
            block: this.getBlock("latest"),
            gasPrice: this.getGasPrice().catch((error) => {
                // @TODO: Why is this now failing on Calaveras?
                //console.log(error);
                return null;
            })
        });

        let maxFeePerGas = null, maxPriorityFeePerGas = null;

        if (block && block.baseFeePerGas) {
            // We may want to compute this more accurately in the future,
            // using the formula "check if the base fee is correct".
            // See: https://eips.ethereum.org/EIPS/eip-1559
            maxPriorityFeePerGas = BigNumber.from("1500000000");
            maxFeePerGas = block.baseFeePerGas.mul(2).add(maxPriorityFeePerGas);
        }

        return { maxFeePerGas, maxPriorityFeePerGas, gasPrice };
    }
```

acenolaza: I see, so we don't really care what the user sets in airseeker.json config `chains.<id>.options.txType` and we just try to always use gasPrice even for EIP1559 chains, right?

src/gas-oracle.ts
acenolaza: Right, I missed that if statement. My bad 🙏🏻

src/gas-oracle.ts
acenolaza: Correct. I was confused 👍🏻

src/gas-oracle.ts
vponline: Yes this was my understanding, but another point I missed is whether we want to specify the `txType` as 0 in update-beacons.ts (if we don't, the transaction may be still sent as `type 2` on some chains (at least on polygon?)).

acenolaza: I think we do want to specify `type=0` (hardcode it probably) but I'm now also wondering why need to have that setting in the config if it has no use 🤔. Should we use `legacy` as default value but keep letting the user change the `txType` via config?

src/gas-oracle.ts
```diff
@@ -0,0 +1,252 @@
...
+export const getGasPrice = async (provider: Provider, gasOracleConfig?: GasOracleConfig) => {
+  // Get gas oracle config for provider
+  const chainOptions = getChainProviderConfig(gasOracleConfig);
+
+  // Fetch and process block data
+  await fetchUpdateBlockData(provider, chainOptions);
+  const gasPrice = getChainProviderGasPrice(provider.chainId, provider.providerName);
+
+  return gasPrice;
+};
```
vponline: I have replaced the interval cycle with this `getGasPrice` method which will start fetching block data (and still re-use from state if it's up-to-date) and finally return a gas price.

vponline:
Fetching 20 blocks (mainnet) is taking 20+ seconds so getting blocks sequentially like this won't do. I'll try out a few options to get the block data in parallel. This will require a provider subscription for use though... :thinking:

vponline:
I have updated the Gas Oracle implementation:

The gas oracle will fetch the latest and latest - 20 blocks and get the specified `percentile` from each block. The latest block's `percentile` will be used if the difference between the two percentiles does not exceed the `gasDeviationThreshold`.

`getGasPrice` will be used as a fallback if the `gasDeviationThreshold` is exceeded, either block has less than `minBlockTransactions` or failing to fetch either block.

The `backupGasPriceGwei` config option will be used if failing to fetch both blocks and the fallback `gasPrice`.

src/check-condition.ts
```diff
@@ -15,12 +15,12 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
...
+  onChainValue: ethers.BigNumber,
   deviationThreshold: number,
   apiValue: ethers.BigNumber
-): Promise<boolean> => {
-  const updateInPercentage = calculateUpdateInPercentage(onChainVale, apiValue);
+) => {
+  const updateInPercentage = calculateUpdateInPercentage(onChainValue, apiValue);
   const threshold = ethers.BigNumber.from(Math.trunc(deviationThreshold * HUNDRED_PERCENT)).div(
     ethers.BigNumber.from(100)
   );
```
vponline, line 26: I realized this is no longer an async function as we no longer read the DapiServerValue inside it so I updated it since I am re-using it for the gas-oracle.

bbenligiray:
I'd recommend something like this
```
"options": {
  "txType": "0",
  "gasOracle": {
    "fallbackGasPrice": {
      "value": 10,
      "unit": "gwei"
    },
    "recommendedGasPriceMultiplier": 1.2,
    "latestGasPriceOptions": {
      "percentile": 60,
      "minTransactionCount": 10,
      "pastToCompareInBlocks": 20,
      "maxDeviationMultiplier": 2
    }
  },
  "fulfillmentGasLimit": 500000
}
```
where the "gas oracle" is something that gives you a legacy gas price. `recommendedGasPriceMultiplier` and `latestGasPriceOptions` are optional.
1. Calculates the 60th percentile from the current block (go to 4 if it has less than 10 transactions)
2. Calculates the 60th percentile from 20 blocks ago (go to 4 if it has less than 10 transactions)
3.  If `gaspricenow/2 < gasprice20blocksago > gaspricenow*2` returns `gaspricenow`
4. Call `getGasPrice()` (go to 6 if it fails)
5. Return the recommended gas price times `recommendedGasPriceMultiplier`
6. Return `fallbackGasPrice`

vponline:
> I'd recommend something like this
> ...
> 1. Calculates the 60th percentile from the current block (go to 4 if it has less than 10 transactions)
> 2. Calculates the 60th percentile from 20 blocks ago (go to 4 if it has less than 10 transactions)
> 3. If `gaspricenow/2 < gasprice20blocksago > gaspricenow*2` returns `gaspricenow`
> 4. Call `getGasPrice()` (go to 6 if it fails)
> 5. Return the recommended gas price times `recommendedGasPriceMultiplier`
> 6. Return `fallbackGasPrice`

I have now updated the gas oracle and tests according to the specs above.

config/airseeker.example.json
```diff
@@ -26,13 +26,27 @@
         }
       },
       "options": {
-        "txType": "eip1559",
+        "txType": "legacy",
```
amarthadan, line 29: So, what was the result regarding the transaction type discussion?

src/gas-prices.ts
```diff
@@ -28,9 +28,12 @@ export type GasTarget = LegacyGasTarget | EIP1559GasTarget;
 export const parsePriorityFee = ({ value, unit }: node.PriorityFee) =>
   ethers.utils.parseUnits(value.toString(), unit ?? 'wei');
 
+export const multiplyGasPrice = (gasPrice: BigNumber, gasPriceMultiplier: number) =>
+  gasPrice.mul(BigNumber.from(Math.round(gasPriceMultiplier * 100))).div(BigNumber.from(100));
+
 export const getLegacyGasPrice = async (
   provider: Provider,
-  chainOptions: node.ChainOptions,
+  chainOptions: node.ChainOptions & { gasPriceMultiplier?: number },
```
amarthadan, line 36: I might be just blind, but I don't see you passing `gasPriceMultiplier` anywhere in the code to the `getGasPrice` function. Actually, I don't see you using our `getGasPrice` (the one from this file, not the one on the RPC provider instance) anywhere. Do we not need this anymore or is that just a mistake?

src/gas-prices.test.ts
```diff
@@ -141,6 +141,35 @@ describe('getGasPrice', () => {
...
+    const stateSpy = jest.spyOn(state, 'getState');
+    stateSpy.mockImplementation(
+      () =>
+        ({
+          config: {
+            chains: {
+              '31337': {
+                options: {
+                  txType: 'legacy',
+                  gasPriceMultiplier,
```
amarthadan: There's no such field in the config file as far as I can tell.

src/gas-prices.test.ts
```diff
@@ -141,6 +141,35 @@ describe('getGasPrice', () => {
     expect(getBlock).toHaveBeenCalledTimes(0);
   });
 
+  it('applies gasPriceMultiplier to non-EIP-1559 provider', async () => {
+    const gasPriceMultiplier = 1.75;
+    const stateSpy = jest.spyOn(state, 'getState');
```
amarthadan: Can't you just add another chain to the `createMockState` output instead of this?

src/gas-prices.ts
```diff
@@ -28,9 +28,12 @@ export type GasTarget = LegacyGasTarget | EIP1559GasTarget;
 export const parsePriorityFee = ({ value, unit }: node.PriorityFee) =>
   ethers.utils.parseUnits(value.toString(), unit ?? 'wei');
 
+export const multiplyGasPrice = (gasPrice: BigNumber, gasPriceMultiplier: number) =>
+  gasPrice.mul(BigNumber.from(Math.round(gasPriceMultiplier * 100))).div(BigNumber.from(100));
```
amarthadan, line 32: The same function is also defined in `gas-oracle.ts`. It should be defined only in one place.

src/utils.ts
```diff
@@ -1,3 +1,17 @@
...
+import { INFINITE_RETRIES, PROVIDER_TIMEOUT_MS, RANDOM_BACKOFF_MAX_MS, RANDOM_BACKOFF_MIN_MS } from './constants';
+
 export const sleep = (ms: number) => new Promise((res) => setTimeout(res, ms));
 
 export const shortenAddress = (address: string) => address.replace(address.substring(5, 38), '...');
+
+export const calculateTimeout = (startTime: number, totalTimeout: number) => totalTimeout - (Date.now() - startTime);
+
+// We retry all chain operations with a random back-off infinitely until the next updates cycle
+// TODO: Errors are not displayed with this approach. Problem?
```
amarthadan: This TODO doesn't apply anymore and can be removed.

test/e2e/airseeker.feature.ts
```diff
@@ -248,62 +248,6 @@ describe('Airseeker', () => {
     expect(beaconValueBTCNew).toEqual(hre.ethers.BigNumber.from(43_000 * 1_000_000));
   });
 
-  it('updates the beacons successfully with one invalid provider present', async () => {
```
amarthadan: Why was this removed?

README.md
```diff
@@ -22,6 +22,25 @@ read from the environment variables. When running locally, either just with `yar
 are automatically loaded from `config/secrets.env` file. Take a look at `config/secrets.example.env` for an example
 secrets file.
 
+### Gas oracle options
+
+- `fallbackGasPrice`: (required) - The gas price to use for beacon update transactions if fetching both blocks and
+  fallback gas prices fails. Defined as an object, e.g. `{"value": 10, "unit": "gwei"}`.
+- `maxTimeout`: (optional) - The maximum timeout (in seconds) for fetching a block or fallback gas price (defaults to
```
amarthadan, line 29: Do we need this configurable? We already have a similar timeout for fetching data from an HTTP gateway and that one is hardcoded.

src/gas-oracle.ts
```diff
@@ -0,0 +1,200 @@
...
+  // Calculate gas price percentiles for each block
+  const blockPercentileGasPrices = resolvedGoBlocks.reduce(
+    (acc: { blockNumber: number; percentileGasPrice: ethers.BigNumber }[], block) => {
+      // Stop processing if fetching the block was not succesful
+      // or if the block does not have enough transactions
+      if (!block.success || block.data.transactions.length < minTransactionCount) return acc;
+
+      const percentileGasPrice = getPercentile(
+        percentile,
+        block.data.transactions.map((tx) => (tx.gasPrice || block.data.baseFeePerGas!.add(tx.maxPriorityFeePerGas!))!)
```
amarthadan: Are these exclamation points ok? Are those values always available? Why are they as optional in ethers?

src/gas-oracle.ts
```diff
@@ -0,0 +1,200 @@
...
+      // should have been passed in at this point
+      if (!percentileGasPrice) return acc;
+
+      return [...acc, { percentileGasPrice, blockNumber: block.data.number }];
+    },
+    []
+  );
+
+  // Check percentileGasPrices only if we have the result from at least two blocks
+  if (blockPercentileGasPrices.length > 1) {
```
amarthadan: Isn't this going to be always 2 (or fewer) blocks? We're doing it only for the latest block and the one in history, don't we?

src/gas-oracle.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    // Check that the latest block percentileGasPrice does not exceed the maxDeviationMultiplier compared to
+    // the reference block to protect against gas price spikes
+    const isWithinDeviationLimit = checkMaxDeviationLimit(
+      sortedBlockPercentileGasPrices[1].percentileGasPrice,
+      sortedBlockPercentileGasPrices[0].percentileGasPrice,
+      maxDeviationMultiplier
+    );
+
+    // Return latest the percentile for the latest block if within the limit
```
amarthadan:
```suggestion
    // Return the percentile for the latest block if within the limit
```

src/gas-oracle.ts
```diff
@@ -0,0 +1,200 @@
...
+  // Check percentileGasPrices only if we have the result from at least two blocks
+  if (blockPercentileGasPrices.length > 1) {
+    // Sort by blockNumber
+    const sortedBlockPercentileGasPrices = blockPercentileGasPrices.sort((a, b) => b.blockNumber - a.blockNumber);
+
+    // Check that the latest block percentileGasPrice does not exceed the maxDeviationMultiplier compared to
+    // the reference block to protect against gas price spikes
+    const isWithinDeviationLimit = checkMaxDeviationLimit(
+      sortedBlockPercentileGasPrices[1].percentileGasPrice,
+      sortedBlockPercentileGasPrices[0].percentileGasPrice,
```
amarthadan: Maybe instead of accessing the values via indices every time we can extract them to variables so it's obvious what those values are. Something like `latestPercentileGasPrice` and `referencePercentileGasPrice`.

src/gas-oracle.ts
```diff
@@ -0,0 +1,200 @@
...
+  // Attempt to get gasPrice as fallback
+  const gasStartTime = Date.now();
+  const gasPrice = await go(() => rpcProvider.getGasPrice(), {
+    ...prepareGoOptions(gasStartTime, totalTimeout),
+    onAttemptError: (goError) =>
+      logger.warn(`Failed attempt to get gas price. Error: ${goError.error}`, logOptionsChainId),
+  });
+
+  if (!gasPrice.success) {
+    // Use the hardcoded back if back up if gasTarget cannot be fetched
```
amarthadan:
```suggestion
    // Use the hardcoded back-up gas price if gas price cannot be fetched
```

src/gas-oracle.ts
```diff
@@ -0,0 +1,200 @@
...
+    logger.warn(
+      `Unable to get fallback gasPrice from provider. Using fallbackGasPrice from config set to ${fallbackGasPrice.value} ${fallbackGasPrice.unit}.`,
+      logOptionsChainId
+    );
+    return parsePriorityFee(fallbackGasPrice);
+  }
+
+  logger.info(`Fallback gas price set to ${ethers.utils.formatUnits(gasPrice.data, 'gwei')} gwei`, logOptionsChainId);
+
+  return recommendedGasPriceMultiplier ? multiplyGasPrice(gasPrice.data, recommendedGasPriceMultiplier) : gasPrice.data;
```
amarthadan: I would reverse this condition just so that the code flow follows the fallback flow better. If we have the gas price from the provider, we use that. And if we don't, we use the hardcoded one.

README.md
```diff
@@ -22,6 +22,25 @@ read from the environment variables. When running locally, either just with `yar
...
+  `3`).
+- `recommendedGasPriceMultiplier`: (optional) - The multiplier to apply to the fallback gas price reported by the
```
amarthadan, line 31: I would add a note that the `recommendedGasPriceMultiplier` will not be used on the `fallbackGasPrice` and so that `fallbackGasPrice` will be used as is in the config file.

amarthadan (changes requested):
I haven't reviewed tests yet as I suspect some changes in functionality. I'll do that once those are sorted out.

config/airseeker.example.json
vponline: It was decided to stick with `strings` in the configs for readability, but to use the legacy type, so I changed this in the `airseeker.example.json` to match.

src/gas-prices.ts
vponline: Sorry about not clarifying this in the description. Since the gas oracle is supposed to always return a (legacy) `gasPrice`, I switched to using the `provider.getGasPrice` method directly instead of using the `gas-prices.ts` (these are no longer used at all, but I did not remove the files in this PR). 

We could still use Airnode's `getGasPrice` because it supports different txTypes if needed, so please let me know and I can make the changes.

src/gas-prices.test.ts
vponline: I didn't add it since it's an optional field and only used in the **gas-prices.ts** `getGasPrice` implementation (and those methods are not used in the gas-oracle).

The gas-oracle has a separate multiplier: `recommendedGasPriceMultiplier`.  This setup is a bit confusing, but allows different multipliers if we want to somehow combine both `gas-prices.ts` and gas-oracle?

src/gas-prices.ts
vponline: I was anticipating that we would delete **gas-prices.ts** entirely but was not sure if that should be done here (I guess so if the gas-oracle implementation is approved)

README.md
vponline: > Do we need this configurable? 

Maybe :smile:. It depends on if we are fine with a 5 second timeout or if that may be too long (within a cycle). I made it configurable just in case.

config/airseeker.example.json
bbenligiray: Weren't we supposed to use `"0"` or `"2"`? Or was that only for Airnode?

src/gas-oracle.ts
vponline: You're right, I wrote it like this in case we needed to add more blocks (there was talk of using 3 + 1 reference block before), but it's an easy change if needed so I will change this check to `=== 2`.

src/gas-oracle.ts
vponline: In practice, it seems that `gasPrice` is always returned on all EIP1559 chains as well, even though ethers says that it would be `null`.

Conversely, ethers says that the EIP1559 values would be null on non-EIP1559 chains.

test/e2e/airseeker.feature.ts
vponline: This should not have been removed, I've added it back.

config/airseeker.example.json
amarthadan: Let's deal with this when adding the strategies feature (for reference https://api3workspace.slack.com/archives/C037NLX1R8T/p1653813343064269), keeping it as it is for now.

src/gas-prices.ts
amarthadan: Ok, let's keep the `gas-price.ts` there, unused, for now and we will revisit it when working on the support of multiple strategies.

src/gas-oracle.ts
amarthadan: I would add some conditions just to be on the safe side. We use a fallback if we don't have the necessary data for the calculation anyway. I just want to avoid potential unnecessary exceptions.

vponline: I've added a check for this (and a few unit tests), please see [here](https://github.com/api3dao/airseeker/pull/75/files#diff-38619592ec5069b9dc3cf986de32e5294ae2b0d5acd47aaf650cbf9463108016R96-R105) and let me know what you think.

vponline:
I believe that covers all revisions since we're leaving the config changes for later, so please re-review :smile:

src/gas-prices.ts
vponline: Ok sounds good, hopefully by then the imported Airnode config type will include `gasPriceMultiplier` as well

src/gas-oracle.ts
```diff
@@ -0,0 +1,210 @@
...
+      // Note: percentileGasPrice should never be undefined as only arrays with items
+      // should have been passed in at this point
+      if (!percentileGasPrice) return acc;
+
+      return [...acc, { percentileGasPrice, blockNumber: block.data.number }];
+    },
+    []
+  );
+
+  // Check percentileGasPrices only if we have the result from at least two blocks
```
amarthadan: A bit misleading comment.

src/gas-oracle.test.ts
```diff
@@ -0,0 +1,663 @@
...
+        'getBlockWithTransactions'
+      );
+      const blockDataMock = [
+        {
+          baseFeePerGas: ethers.BigNumber.from(20),
+          number: 23,
+          transactions: [
+            { maxPriorityFeePerGas: ethers.BigNumber.from(2) },
+            { maxPriorityFeePerGas: ethers.BigNumber.from(2) },
+          ],
```
amarthadan: Won't this fall under the `not enough transactions` check instead of the `can't compute gas price from the eip1559 transaction` check?

src/gas-oracle.ts
```diff
@@ -0,0 +1,210 @@
...
+    onAttemptError: (goError) =>
+      logger.warn(`Failed attempt to get gas price. Error: ${goError.error}`, logOptionsChainId),
+  });
+
+  if (gasPrice.success) {
+    logger.info(`Fallback gas price set to ${ethers.utils.formatUnits(gasPrice.data, 'gwei')} gwei`, logOptionsChainId);
+
+    return recommendedGasPriceMultiplier
+      ? multiplyGasPrice(gasPrice.data, recommendedGasPriceMultiplier)
+      : gasPrice.data;
```
amarthadan: Maybe we can use the multiplier before logging so we log the price we're actually using.

src/gas-oracle.test.ts
```diff
@@ -0,0 +1,663 @@
...
+
+      const gasPrice = await api.getOracleGasPrice(stateProvider, defaultGasOracleConfig);
+
+      expect(getBlockWithTransactionsSpy).toHaveBeenNthCalledWith(1, 'latest');
+      expect(getBlockWithTransactionsSpy).toHaveBeenNthCalledWith(2, -20);
+
+      expect(gasPrice).toEqual(prices.parsePriorityFee(defaultGasOracleConfig.fallbackGasPrice));
+    });
+
+    it('returns config backupGasPriceGwei gas if failing to fetch both blocks and fallback', async () => {
```
amarthadan: I think there's no `backupGasPriceGwei` field, right? Is this the old name? Can you update it?

test/e2e/gas-oracle.feature.ts
```diff
@@ -0,0 +1,172 @@
...
+        const gasPrice = await gasOracle.getOracleGasPrice(
+          { ...provider, providerName },
+          gasOracleConfig as GasOracleConfig
+        );
+        const fallbackGasPrice = await provider.rpcProvider.getGasPrice();
+
+        expect(gasPrice).toEqual(fallbackGasPrice);
+      });
+
+      it('uses fallback backupGasPriceGwei when getBlockWithTransactions and getGasPrice provider calls fail', async () => {
```
amarthadan: I think there's no `backupGasPriceGwei` field, right? Is this the old name? Can you update it?

amarthadan (approved):
Ok, some cosmetic nits but other than that I think it's good to go :+1:

test/e2e/gas-oracle.feature.ts
vponline: Yep, nice catch, will fix!

src/gas-oracle.test.ts
vponline: You're right, will fix

config/airseeker.example.json
```diff
@@ -26,13 +26,27 @@
...
         "priorityFee": {
           "value": 3.12,
           "unit": "gwei"
         },
         "baseFeeMultiplier": 2,
```
bbenligiray: These probably shouldn't be in the same object as `gasOracle`

vponline: Do you mean that the `gasOracle` object should not be in `<chainId>.options`? `priorityFee` and `baseFeeMultiplier` will not be used currently so should we leave moving these fields for the config refactor task?

bbenligiray: Okay, I just meant that unused fields should be removed

vponline: Got it,I'll remove them since we are using `txType` 'legacy' anyway.

src/update-beacons.test.ts
acenolaza: Just double checking. The decision is to keep these 2 tests here, correct?

acenolaza (approved):
Reviewed things that I had pointed out and all of them got resolved or answered. LGTM 👍🏻

src/update-beacons.test.ts
vponline: I think it's a good idea to move them as well, but perhaps we can leave that for a later task (e.g. with the config changes)?

PR #76, opened 2022-05-05
BEC-307: Add more thorough validation for configuration file

acenolaza:
Pretty self explanatory 😁 but basically moved validations related to the config that were spread around the code to the validations.ts file and used zod refinements to validate those checks on startup.


test/e2e/airseeker.feature.ts
```diff
@@ -249,75 +249,6 @@ describe('Airseeker', () => {
     expect(beaconValueBTCNew).toEqual(hre.ethers.BigNumber.from(43_000 * 1_000_000));
   });
 
-  it('updates the beacon successfully with one invalid beacon present', async () => {
-    mockReadFileSync(
```
vponline, line 308: This test is removed because the validator won't let an invalid beacon in the config to pass, but I'm wondering if we should still have tests where we make sure that the Airseeker will run correctly for valid beacons even if an invalid one somehow made it to deployment?

This would require mocking the `loadConfig` function instead of mocking the `fs` function (so that the validator is essentially bypassed for this test, we did it like this in [Airkeeper](https://github.com/api3dao/airkeeper/blob/main/test/e2e/psp.feature.ts#L270)).

Just a thought, WDYT?

acenolaza: I had that same exact concern when removing these tests (which is always scary thing to do 😁) but since the validations no longer exist in the code that process the triggers then the code will probably fail or even crash if the config has invalid values. That is also why I added quite a few tests to the validation.test.ts file in order to make sure that the app will not start if it has any invalid values in the config.

An alternative could be to have replace these tests with other tests that expect the app to fail or crash if there are invalid values in the config but that is not something we really expect I guess.

vponline: Thanks for explaining. I guess there is no way for a `beaconId` to suddenly go `invalid` while the Airseeker is running so perhaps these are not necessary :+1:

src/providers.test.ts
```diff
@@ -48,7 +48,6 @@ describe('initializeProviders', () => {
     triggers: {
       beaconUpdates: {
         '1': {},
-        '2': {},
```
amarthadan, line 51: Why is ok to remove this without changing anything else?

src/providers.ts
```diff
@@ -10,16 +9,10 @@ export const initializeProvider = (chainId: string, providerUrl: string): Omit<P
 
 export const initializeProviders = () => {
   const { config } = getState();
-  const chains = Object.keys(config.triggers.beaconUpdates);
-  const providers = chains.reduce((acc: Providers, chainId: string) => {
+  const sponsorBeaconsByChainId = Object.keys(config.triggers.beaconUpdates);
```
amarthadan: The name `sponsorBeaconsByChainId` suggests that the variable has a structure like:
```
{
  <chain ID>: <beacon object>,
  ...
}
```
which is not true. I think calling it `chains` is fine.

src/validation.ts
```diff
@@ -82,6 +86,68 @@ export const triggersSchema = z.object({
...
+    if (derivedBeaconId !== beaconId) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: `Beacon ID "${beaconId}" is invalid`,
+        path: [beaconId],
+      });
+    }
+
+    // Verify TemplateId
+    const template = config.templates[beacon.templateId];
```
amarthadan, line 136: If you go just through the referenced templates you can have a template in `templates` that is not referenced in `beacons` and it would pass. I don't think we want that, do we?

src/validation.ts
```diff
@@ -93,8 +159,9 @@ export const configSchema = z
     templates: templatesSchema,
     triggers: triggersSchema,
   })
-  .strict();
-
+  .strict()
+  .superRefine(validateBeacons)
+  .superRefine(validateTriggersBeaconUpdates);
```
amarthadan: I think the names are a bit confusing. For example, you're calling the function `validateBeacons` but you're also validating template IDs. I would probably go with a separate function for each validation, something like `validateBeaconIDs`, `validateTemplateIDs`, `validateBeaconTemplateReferences`, etc. WDYT?

Also, can you add a validation that the referenced gateway exists (the `airnode` in `beacons` and a key in `gateways`)?

src/validation.test.ts
```diff
@@ -18,3 +20,158 @@ it('successfully parses example configuration', () => {
...
+      {
+        code: 'custom',
+        message: `Beacon ID "${firstBeaconId}" is not defined in the config.beacons object`,
+        path: ['triggers', 'beaconUpdates', '3', '0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC', 'beacons', 0],
+      },
+    ])
+  );
+});
+
+it('fails if derived beaconId is different to beacons.beaconId', () => {
```
amarthadan: Is this title correct? You're changing the template ID, not beacon ID.

src/validation.test.ts
```diff
@@ -18,3 +20,158 @@ it('successfully parses example configuration', () => {
...
+  config.beacons[firstBeaconId].templateId = randomTemplateId;
+  const interpolatedConfig = interpolateSecrets(config, envVariables);
+
+  expect(() => configSchema.parse(interpolatedConfig)).toThrow(
+    new ZodError([
+      {
+        code: 'custom',
+        message: `Beacon ID "${firstBeaconId}" is invalid`,
+        path: [firstBeaconId],
+      },
```
amarthadan: Why is there a message that beacon ID is invalid?

amarthadan (changes requested):
I very much like this PR. I think both code and tests are way cleaner without the validation and all the validation is in one place :+1:

src/providers.test.ts
acenolaza: Because previous to these changes the validation in the code would handle that trigger and just ignore it. Now that's no longer being handled in the code and the test is not validating the config therefore test will fail. That value should never be there if the config validation is run during startup tho.

src/providers.ts
acenolaza: Yes, the new name is not super accurate but I thought that chains could be misleading to the one reading the code since there is a top level [chains object](https://github.com/api3dao/airseeker/blob/bec307-custom-config-validation/config/airseeker.example.json#L15) in the config file.

I don't feel this change is too meaningful so I can revert it back to `chains` or use a different one. Maybe `beaconUpdatesChains`? Naming things is really hard 😅

src/validation.ts
acenolaza: Well that wasn't being validated even before these changes because the app perform actions based on the items under `beacons` (triggers). 

For instance in Airkeeper I usually had many test templates and endpoints that were not being referenced in the triggers at one point but while testing I would switch between them (referencing ones and stop referencing others).

Bottom line is that since we are iterating over the triggers then all things referenced by them must exist but I'm not sure if we should also check for things not being in used by the triggers.

src/validation.ts
acenolaza: I called them like that mainly based on the thing I'm iterating over:
```
validateBeacons -> config.beacons
validateTriggerBeaconUpdates -> config.triggers.beaconUpdates
```
I don't mind using the names you are suggesting but that would require looping over the same `config.beacons` list more than once. Would that be OK you think?

src/validation.test.ts
acenolaza: I think you make a good point here. I was changing one of the components used when computing the `beaconId` hash but the test might be more accurate if I change the `beaconId` instead.

src/validation.test.ts
acenolaza: Basically because of [this](https://github.com/api3dao/airseeker/pull/76#discussion_r869414267). 
`BeaconID = hash(airnode, templateId)`

So to try to make things more clear:
1. this test "fails if beacons.beaconId in beacons is invalid" checks that `beacons.beaconId` referenced in `triggers` exist.
2. this test "fails if derived beaconId is different to beacons.beaconId" checks that the `beaconId` key in `beacons` matches the derived hash of it's children entries (`airnode` and `templateId`)

acenolaza: Disregard my previous comment. It turns out one of those tests were redundant so ended up removing it.

src/providers.ts
amarthadan: Yeah, `beaconUpdatesChains` sounds good.

src/validation.ts
amarthadan: Ok. I thought that we would allow only what's referenced and if something is not referenced it would be considered an issue. But I understand your reasoning, it makes sense to allow unused bits that are not referenced.

src/validation.ts
amarthadan: I understand and think that would be fine. I mainly suggested the names just so you can tell just by glancing at it what's getting validated. But if you think that it would be a lot slower and it doesn't really bring much value feel free to keep it as it is :slightly_smiling_face:

amarthadan (changes requested):
Can you add a validation that the referenced gateway exists (the `airnode` in `beacons` and a key in `gateways`)?

src/validation.ts
acenolaza: I think you'll be pleased with [this](https://github.com/api3dao/airseeker/commit/a2b76debb4aa35620b1edad55728b0864cac18b6) last commit 🤞🏻 😅  

Now I just need to add the gateway reference check 💪🏻

acenolaza: Done!

acenolaza:
> Can you add a validation that the referenced gateway exists (the `airnode` in `beacons` and a key in `gateways`)?

Done! https://github.com/api3dao/airseeker/pull/76/commits/b4bbf3e82ece36745dc1e41ae1e09e6d609dcdc4

src/validation.ts
```diff
@@ -23,14 +25,29 @@ export const beaconSchema = z
...
+    const derivedBeaconId = ethers.utils.solidityKeccak256(['address', 'bytes32'], [beacon.airnode, beacon.templateId]);
+    if (derivedBeaconId !== beaconId) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: `Beacon ID "${beaconId}" is invalid`,
+        path: [beaconId],
+      });
+    }
+  });
+});
```
amarthadan, line 41: You were right, I very much like this :smile:

amarthadan (approved):
Awesome :+1:

PR #91, opened 2022-05-12
Add Serverless Lambda Deployment Ability

aquarat:
Not sure if this should be merged :shrug: 

These changes enabled Airseeker to run on Lambda... and it's working pretty much perfectly in production.

Running Airseeker like this will cost ~$12 per month in total (arm64 ftw)


aws-lambda-deploy-guide.md
```diff
@@ -0,0 +1,24 @@
...
+
+Export AWS credentials in your terminal or ensure your local aws installation is logged in.
+
+Run:
+
+```shell
+yarn sls deploy --region us-east-1
+```
+
+The app won't immediately start (due to the scheduler). You can jumpstart the app by calling
```
amarthadan: That will potentially cause two Airseeker runs to overlap for up to 13 minutes, right? Is that ok?

aws-lambda-deploy-guide.md
```diff
@@ -0,0 +1,24 @@
...
+
+```shell
+yarn sls invoke --function airseeker &
```
amarthadan, line 18: This will invoke the lambda on AWS, not locally, right? Why run it in the background?

src/main-handler.ts
```diff
@@ -0,0 +1,4 @@
+// This file is used when running Airseeker directly via NodeJS
+import { main } from './main';
+
+main();
```
amarthadan: Maybe I'm missing something. The content of this file is the same as was the content of `index.ts`. Why can't we just use the `index.ts`?

src/main.ts
```diff
@@ -17,11 +17,22 @@ export async function main() {
...
-  process.on('SIGTERM', handleStopSignal);
+/**
+ * This function is required to make the process block because loop promises are not returned/not hierarchical.
+ * We need a separate function from main to keep Jest happy (without having a dedicated Jest check).
+ */
+export async function serverlessHandler() {
+  await main();
+
+  await new Promise((_resolve) => {});
 }
```
amarthadan: I would propose moving this function to a separate file, something like `serverless.ts` as it's an environment-specific "hack" and not a main functionality.

src/main.ts
```diff
@@ -17,11 +17,22 @@ export async function main() {
...
-  process.on('SIGINT', handleStopSignal);
-  process.on('SIGTERM', handleStopSignal);
+/**
+ * This function is required to make the process block because loop promises are not returned/not hierarchical.
+ * We need a separate function from main to keep Jest happy (without having a dedicated Jest check).
+ */
+export async function serverlessHandler() {
+  await main();
+
+  await new Promise((_resolve) => {});
```
amarthadan: Can you explain this a bit more? AFAIK even though the `main` function finishes, node waits for any unfinished async functions. That's probably not the case with lambdas, right? How does this promise help with that? It resolves immediately, doesn't it?

amarthadan (changes requested):

src/main.ts
Siegrift: >  It resolves immediately, doesn't it?

This one never resolves. (And I'll let @aquarat answer the other questions).

amarthadan: So, does it (a)waits forever and we're just waiting for AWS to kill it?

aws-lambda-deploy-guide.md
aquarat: I'm not sure when the scheduler is triggered initially? I'm struggling to find documentation for it.
If they overlap, the worst thing is that we'll have more provider and RPC calls, more API calls and we'll pay a little more for the runtime :shrug:

aws-lambda-deploy-guide.md
aquarat: I should probably add something extra here:
This invokes the cloud hosted Lambda, which won't return for 15 minutes, so the `&` lets you use the terminal until the command dies or times out.
A better approach might be to store the PID of the sls process, wait 10 seconds and then kill that PID.

src/main-handler.ts
aquarat: Lambda doesn't execute a file, it calls a function, so the working function must be exported in a file that does not have the `main()` call. If you call `main()` in index.ts it'll execute when you try and import the package.

src/main.ts
aquarat: Fine with me :)

aws-lambda-deploy-guide.md
amarthadan: I think you normally provide a schedule with a cron job syntax and this is just an abbreviation by serverless. So, I would guess something like `*/13 * * * *`.

src/main.ts
aquarat: > That's probably not the case with lambdas, right? How does this promise help with that? It resolves immediately, doesn't it?

AFAIK you are correct: Node won't quit if you have waiting callbacks, which is why this application works correctly in a "conventional" environment. On Lambda, however, the runtime will "freeze" if you don't explicitly handle promise resolution in the handler function. What's weirder is that dangling `async` calls will resume executing when the Lambda is invoked again (thawing) (assuming the context stays warm, max 2.5 hours).

The timeout blocks indefinitely until AWS kills it. You have graceful signal handling in Airseeker, which is pleasing to see, it's elegant and _correct_ but it isn't necessary imo/AFAIK because you don't have a dirty state that needs to be cleaned up; if the Node process quits all associated resources will be garbage collected. It's not like a database that has to flush its writes to disk before quitting for consistency. Therefore having AWS kill the runtime on timeout is an efficient approach - and I suspect it'd be adequate in conventional Node too, aka `process.exit(0)`;

aquarat:
As always, thanks for the comments, I'll make the changes soon :)

aws-lambda-deploy-guide.md
amarthadan: Oh, I thought that it will just start the lambda but the prompt is returned to you. But it still runs on AWS and it logs to CloudWatch, right? What you see is just a copy of STDOUT? Would it even stop the process if you would send SIGINT (Ctrl-C)?

src/main-handler.ts
amarthadan: I understand that but where are you importing `index`? Lambda will execute `src/main.serverlessHandler`, which has nothing to do with it. The only place you use `main-hanlder.ts` is the pm2 configuration and you're just replacing the `index.js` there.

src/main.ts
amarthadan: The signal handling is fine, the "graceful" part is unnecessary and actually quite annoying and should be removed. Sorry, that's on me :see_no_evil:

Two concerns:
* Does the lambda clean its context if the lambda itself kills the processing? Won't we have two (or more) Ariseekers running in the next lambda?
*   > and I suspect it'd be adequate in conventional Node too, aka process.exit(0)
  
    I actually doubt that, I'd say it will be something like `kill -9 <PID>` (not that it matters that much). But my concern is that every lambda will technically fail, right? From the AWS's point of view. It will be also logged that way if I'm not mistaken.

aws-lambda-deploy-guide.md
aquarat: That's what I thought @ cron expression... but it seems that it might be based on the creation time of the scheduler. It needs testing/research.

aws-lambda-deploy-guide.md
aquarat: It is not a copy of stdout. Lambda expects handlers to return an HTTP response object AFAIK, like this:
```
export const chainValueDataPoint: APIGatewayProxyHandler = async (event): Promise<any> => {
  return {
    statusCode: 200,
    headers: {something: 'something'},
    body: JSON.stringify({time: Date.now()}),
  };
```
The function never returns so you would not see anything until it times out and then it would be an error message.

I don't know for sure if terminating the command early would also terminate the Lambda.

src/main-handler.ts
aquarat: I ran into this problem when I tried deploying to GCP. GCP Cloud Functions rely on the `package.json` defined `main` field.
This will be a problem if we try importing resources from Airseeker in other projects via the index. Anything that relies on the `main` field in `package.json` will call `main()` on import.

We need a file that acts as the index entry point and another file that acts as a `main` entry point for applications that use Airseeker as a library vs direct execution. The files can be named differently.

src/main.ts
aquarat: > Sorry, that's on me see_no_evil

I thought it was cool :)

> Ariseekers running in the next lambda?

That's an interesting point, I will check the logs but it seems to kill off everything on each cycle - for instance the RAM usage per cycle is fairly consistent at between 160 and 180 MB. I also can't see any indication of overlapping loops in the logs; log entries for specific concerns come in waves - so perhaps AWS completely resets the environment on error.

> From the AWS's point of view. It will be also logged that way if I'm not mistaken.

You're right, looking at the Lambda metrics graphs AWS regards every run as an error. The status message in the logs is "Task timed out after 900.11 seconds"

aquarat: It actually looks like it retries executing the Lambda on an error, so we effectively have two Lambdas running concurrently (!) but in different execution contexts. Thanks for querying this.

aws-lambda-deploy-guide.md
amarthadan: > The function never returns so you would not see anything until it times out and then it would be an error message.

I understand why but that's quite an odd behavior. It would be nice if you could document that.

> I don't know for sure if terminating the command early would also terminate the Lambda.

I don't think it does but it's worth checking. Imagine a situation where someone runs this, it starts a process in the background and the person turns their PC off. The system will try to gracefully kill the process (which is, if I'm not mistaken, pretty much sending SIGINT/SIGTERM) and lambda would stop. I know it's a bit of a stretch but the process is running for about 15 minutes and that's enough time to forget about some process in the background. This behavior should be IMO properly documented.

I would also mention that this WILL NOT start the Airseeker "cycling", once this call is done the Airseeker (that one lambda) stops and that's it. So, I'm not sure if "jumpstart" is the right word.

src/main.ts
amarthadan: > so perhaps AWS completely resets the environment on error.

:man_facepalming: It does. I haven't realized it's actually an error even though I wrote it a bulletpoint later :joy: Yes, every time I re-tried lambda from an error state it was basically a cold-start so I presume it does.

> You're right, looking at the Lambda metrics graphs AWS regards every run as an error. The status message in the logs is "Task timed out after 900.11 seconds"

That's quite unfortunate (and by unfortunate I mean ugly :joy:) We're technically giving up any chance of using AWS alerting or anything like that. And we will be doing the same for GCP and any other cloud provider for that matter. If we would want to check any problems in functions run (on the cloud provider layer I mean) we won't be able to filter them efficiently as they will be all failed.

amarthadan: > It actually looks like it retries executing the Lambda on an error, so we effectively have two Lambdas running concurrently (!) but in different execution contexts. Thanks for querying this.

I think that retry behavior is configurable.

aquarat: > I think that retry behavior is configurable.

Yes it is, I have just added a commit to fix this :) Thanks again for leading me down this path.

> That's quite unfortunate (and by unfortunate I mean ugly) 

The benefit of the error scenario is we start fresh every 15 minutes it seems, which removes any leftover state, but I get what you're saying, definitely not ideal. I wonder if we could call `process.exit(0)` in a Lambda? Or enumerate all callbacks globally and destroy them. I ask because we could do this on a timer perhaps and gracefully exit before the timeout.

src/main-handler.ts
amarthadan: I see. But don't you need the same blocking behavior as in `serverless.ts` for GCP as well? Because in this setup GCP will access `main.js` (as stated in `package.json`) exporting the `main` function and you can run just that. You won't be able to access `handler` function from `serverless.ts`.

src/main.ts
amarthadan: I think that any effort to improve this would be just a waste of time and would result in a more complex code. Also, we kinda rely on the state being cleaned and that happens only for failed lambdas. So, I don't think we can do much here. Lambdas are just not designed to be used this way.

aws-lambda-deploy-guide.md
aquarat: I've just tested and if you kill the local invoke the target Lambda doesn't stop.

I'll update the documentation to reflect the above nuances.

src/main-handler.ts
aquarat: `index.ts` should ideally export everything exportable, I'll modify it to do that assuming you agree.

src/main.ts
aquarat: I think that's a bit unfair.

My goal was to quickly adapt Airseeker to Lambda and that has been successful. Airseeker with minimal additional code runs largely perfectly on Lambda. It's cheap, it's reliable and it's very simple to deploy and tear down. The only deficiency as it currently stands is the termination of the handler and that's a product of the application spawning untracked promises and relying on a global state for signalling termination. This isn't an inherently incorrect approach, it's what you went with as ECS as a target.

The bulk of the work of adding serverless and modifying the entry point files took half a day, which is a great return on time invested. It's this discussion that's taken much longer.

No one is suggesting that your work be thrown out or heavily modified. The objective is co-existence, which is what this PR delivers.

amarthadan: I'm not dissing the solution or its time/value efficiency. I'm just saying that Lambdas are not designed to live for a long time and when they timeout they technically fail with an error. Those are facts. And if you want to take advantage of the state being cleaned you need to let the lambda fail, otherwise the state will be frozen and used for another run, right? I'm just making sure we're aware of the tradeoffs, that's all. Sorry, if I sounded harsh, it wasn't meant to be that way.

src/main-handler.ts
amarthadan: > `index.ts` should ideally export everything exportable, I'll modify it to do that assuming you agree.

I'm fine with that but the question still stands. Don't you need the blocking behavior for GCP as well?

aws-lambda-deploy-guide.md
aquarat: I have added the detail to better describe what's going on, LMKWYT please :)

amarthadan: Sounds good :slightly_smiling_face:

src/main-handler.ts
aquarat: I doesn't look like GCP requires the blocking behaviour (although I included it): https://github.com/api3dao/airseeker/blob/add-gcp/src/main.ts

But GCP does require an HTTP server for health checks.

I'm getting a little lost with what you'd like. From what I can see we need:
- an index file (package.json->main referenced) which exports everything in Airseeker without running anything.
- A direct execution entry point (was in `index.ts`, now is in `main-handler.ts`, probably a bad name, maybe it should be called `entry-point.ts`
- The _main_ application logic (currently exported as `main` in `main.ts`, perhaps should be named `airseeker.ts`)
- Serverless handler stub file (`serverless.ts`), will contain a handler for GCP and a handler for Lambda (they have slightly different requirements)

Is it a file naming issue? Let me know if you'd prefer to chat about it quick.

serverless.aws.yml
```diff
@@ -0,0 +1,27 @@
...
+  runtime: nodejs14.x
+  architecture: arm64
+  logRetentionInDays: 14
+  environment: ${file(load-secrets-env.js)}
+
+functions:
+  airseeker:
+    handler: src/serverless.handler
+    maximumRetryAttempts: 0
+    maximumEventAge: 60
```
amarthadan: This doesn't do anything if we have no retries, does it?

src/main.ts
aquarat: No problem :pray: I appreciate it :)

You are correct^
Lambdas are not designed to live for a long time, where a long time is longer than the timeout. This varies greatly between providers, for instance GCP can do an hour timeout on [Gen 2 Cloud Functions](https://cloud.google.com/functions/docs/2nd-gen/overview#:~:text=For%20event%2Dtriggered%20functions%2C%20the,intensive%20and%20more%20parallel%20workloads) and Azure has no timeout. GCP and Azure are very close to ECS in behaviour from what I can tell.

I think the 15 minute interval and associated state clean isn't a bad thing. I know you didn't architect the app to work like this, but it does work and it works well - but again, it's one of multiple solutions that should co-exist :)

src/main-handler.ts
amarthadan: I was just trying to understand what will be used where and how. Let's rename `main-handler.ts` to `entrypoint.ts` and I think it's fine.

serverless.aws.yml
aquarat: You're right, I'll remove it. Paranoia.

src/main-handler.ts
aquarat: Awesome, thank you :)

amarthadan (approved):

PR #101, opened 2022-05-17
Silence logger in tests

Siegrift:
https://api3dao.atlassian.net/browse/BEC-356

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,8 +24,8 @@ jobs:
         with:
           config-file: .github/workflows/mlc_config.json
 
-  lint-build-test:
-    name: Lint, build, test
+  lint-build:
+    name: Lint, build
     runs-on: ubuntu-latest
     # Don't run twice for a push within an internal PR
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -24,8 +24,8 @@ jobs:
         with:
           config-file: .github/workflows/mlc_config.json
 
-  lint-build-test:
-    name: Lint, build, test
+  lint-build:
```
amarthadan, line 27: Is this division actually faster? I vaguely remember asking @vponline about this some time ago and he said that keeping it together is faster so we decided to keep it that way.

amarthadan (approved):

.github/workflows/main.yml
Siegrift: It is not faster, but there are other advantages:
- you can only rerun failed checks in the CI... So for example, when the linter fails, but all other checks pass you can just rerun that and merge more quickly
- As soon as one of the checks fail, no other check is run. If they are separate you can see multiple errors at the same time (e.g. linter failed, e2e test failed).

amarthadan: Fair enough :+1:

PR #102, opened 2022-05-17
Fix ts configuration

Siegrift:
https://api3dao.atlassian.net/browse/BEC-311

The work is inspired by:
- https://github.com/RyanCavanaugh/project-references-demo
- https://www.typescriptlang.org/docs/handbook/project-references.html

### DIFF
--- a/scripts/tsconfig.json
+++ b/scripts/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "extends": "../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "./dist"
+  }
+}
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,7 +2,8 @@
 
 # Package directories & files
 /node_modules
-/dist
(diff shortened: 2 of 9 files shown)


scripts/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "./dist"
+  }
+}
```
amarthadan, line 7: With this we can't import anything from `src`, right? I guess that's not a problem just making sure I understand it correctly.

amarthadan (approved):
I actually quite like this approach. Way better than the build/test switching in `package.json` you've originally suggested.

scripts/tsconfig.json
Siegrift: Yes. All source files (including imports) must be under root dir. So consider the scripts folder as a standalone project independent of test and src.

PR #103, opened 2022-05-23
Switch Serverless to pre-built files

aquarat:
This minor PR removes the serverless typescript plugin and makes Serverless use the `yarn build` output as an execution target.

There's also some extra detail in the deployment guide.

### DIFF
--- a/aws-lambda-deploy-guide.md
+++ b/aws-lambda-deploy-guide.md
@@ -8,18 +8,20 @@ Export AWS credentials in your terminal or ensure your local aws installation is
 yarn build
 ```
 
-### Airseeker Deployment Strategies
+### General Airseeker Deployment Notes
 
 The deployed Airseeker function won't immediately start (due to the scheduler).
 
-There are two main strategies for dealing with this:
+There are three main strategies for dealing with this:
 
-1. Deploy this Airseeker using a different service name in `serverless.yaml`, so that both Airseekers can run
-   concurrently. Wait for the new Airseeker to start. Once it does, ensure that you're satisfied that it is functioning
-   correctly (by referring to the target chains and/or CloudWatch logs). You may then revert the name of the service
-   stack to the old Airseeker and remove it or alternatively delete it using the AWS CloudFormation console.
+1. If an Airseeker is already deployed, deploy the new Airseeker using a different service name in `serverless.yaml`, so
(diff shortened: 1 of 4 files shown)


aws-lambda-deploy-guide.md
```diff
@@ -2,14 +2,32 @@
...
+
+There are two main strategies for dealing with this:
+
+1. Deploy this Airseeker using a different service name in `serverless.yaml`, so that both Airseekers can run
+   concurrently. Wait for the new Airseeker to start. Once it does, ensure that you're satisfied that it is functioning
+   correctly (by referring to the target chains and/or CloudWatch logs). You may then revert the name of the service
+   stack to the old Airseeker and remove it or alternatively delete it using the AWS CloudFormation console.
+2. Deploy Airseeker using the default service name and start the new Airseeker immediately using the invoke commands
+   below.
+
```
amarthadan: I don't think point 1. has anything to do with Airseeker function not starting immediately. I guess what you're describing there is how would you "update" an Airseeker deployment?

amarthadan (approved):

aws-lambda-deploy-guide.md
aquarat: ah yes, you're right. I find it's nice to run both the old and new Airseekers temporarily concurrently, check for correct functioning and then only remove the old Airseeker. It makes for a stable transition and gives one a chance to identify problems.

amarthadan: I agree with that but that should imo go into a different section.

acenolaza: Just wondering, could this cause any race condition when updating if both are using the same sponsor wallet?

aquarat:
I forgot about this PR :disappointed: 
It has to be added to the pending release as the changes that were made to the `tsconfig.json` files broke the Serverless Typescript build and therefore Serverless must use the normal build output.
I'll merge this as-is. I don't believe the issues raised above are sufficiently problematic to block merging.

aws-lambda-deploy-guide.md
aquarat: Better late than never: There would be a race condition, yes, but there is a series of race conditions every 14 minutes when the previous invocation and the new invocation overlap. The race conditions don't seem to be a problem (in terms of committed transactions) in practice.

PR #144, opened 2022-06-15
Enable Docker BuildKit for CI

amarthadan:
Not really solving the auth issue but by enabling BuildKit we skip those unnecessary build stages :man_shrugging:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,7 +24,8 @@ COPY . ${buildDir}
 FROM preparation as sourcegit
 
 ARG branch=main
-ARG repository=https://github.com/api3dao/airseeker.git
+# TODO: Most likely won't work, needs credentials
+ARG repository=https://github.com/API3-CTT/airseeker
 
 RUN git clone --single-branch --branch ${branch} ${repository} ${buildDir}
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   TARGET_NODE_VERSION: '14.17.1'
+  DOCKER_BUILDKIT: '1'
 
 jobs:
   documentation:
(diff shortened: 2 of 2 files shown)


docker/Dockerfile
```diff
@@ -24,7 +24,8 @@ COPY . ${buildDir}
 FROM preparation as sourcegit
 
 ARG branch=main
-ARG repository=https://github.com/api3dao/airseeker.git
+# TODO: Most likely won't work, needs credentials
```
Siegrift, line 27: But I see the CI check pass, what do you mean by `this won't work`?

amarthadan: That's what I said in the description. This stage is skipped because we're using build type `local` by default. The Dockerfile is built in a way that you can build either from your local directory - the `local` type or from Git repository - the `git` type, specifying repository and branch/commit you want to build. Without BildKit enabled, Docker can't skip the build step even though is not needed, so, by enabling it we're skipping the step that would not work. What I meant by the comment is that if you try running the `git` type build it most likely won't work because of the missing credentials.

aquarat (approved):
I don't mind the local build approach.
I think we're also missing Docker Hub credentials for pushing - I can't add these currently because I'm not an admin. I'll ask @bbenligiray

amarthadan:
> I don't mind the local build approach. I think we're also missing Docker Hub credentials for pushing - I can't add these currently because I'm not an admin. I'll ask @bbenligiray

Ok, I'll merge it once the credentials are added.

bbenligiray:
Did we not stop releasing public pre-built images?

aquarat:
> Did we not stop releasing public pre-built images?

The Docker Hub repository is private now - but you have a point, probably no reason to build images.

amarthadan:
So, what's the verdict? Are you going to add the DockerHub credentials or should I drop the Docker builds altogether? I would like to resolve this ASAP, all the PRs are failing.

aquarat:
Let's drop the Docker builds.
It's easy to build this locally. Docker builds are mostly for public use anyway and private Docker Hub repos require space management.

amarthadan:
Ok, I'm closing this and will open a new PR dropping the Docker builds.

PR #150, opened 2022-06-20
Drop Docker builds in CI

amarthadan:
As agreed in https://github.com/API3-CTT/airseeker/pull/144


Siegrift (approved):
LGTM, @aquarat can you also double check? I'd like to merge this since all CI checks fail at the moment.

aquarat (approved):
LGTM too, thanks @amarthadan :smile:

PR #151, opened 2022-06-23
Add check for missing block data to gas-oracle

vponline:
This fixes a bug where `block` is `null` (or `undefined`) which causes the gas oracle to fail with an unexpected error stopping the Airseeker and adds tests for this case. 

```
{
    "errorType": "Runtime.UnhandledPromiseRejection",
    "errorMessage": "TypeError: Cannot read property 'transactions' of null",
    "reason": {
        "errorType": "TypeError",
        "errorMessage": "Cannot read property 'transactions' of null",
        "stack": [
            "TypeError: Cannot read property 'transactions' of null",
            "    at /var/task/dist/gas-oracle.js:118:54",
            "    at Array.reduce (<anonymous>)",
            "    at /var/task/dist/gas-oracle.js:115:61",
            "    at step (/var/task/dist/gas-oracle.js:44:23)",
            "    at Object.next (/var/task/dist/gas-oracle.js:25:53)",
            "    at fulfilled (/var/task/dist/gas-oracle.js:16:58)",
            "    at processTicksAndRejections (internal/process/task_queues.js:95:5)"
        ]
    },
    "promise": {},
    "stack": [
        "Runtime.UnhandledPromiseRejection: TypeError: Cannot read property 'transactions' of null",
        "    at process.<anonymous> (/var/runtime/index.js:35:15)",
        "    at process.emit (events.js:400:28)",
        "    at process.emit (domain.js:475:12)",
        "    at processPromiseRejections (internal/process/promises.js:245:33)",
        "    at processTicksAndRejections (internal/process/task_queues.js:96:32)"
    ]
}
```


src/gas-oracle.ts
```diff
@@ -89,9 +89,15 @@ export const fetchBlockData = async (provider: Provider, gasOracleOptions: GasOr
   // Calculate gas price percentiles for each block
   const blockPercentileGasPrices = resolvedGoBlocks.reduce(
     (acc: { blockNumber: number; percentileGasPrice: ethers.BigNumber }[], block) => {
-      // Stop processing if fetching the block was not succesful
+      // Stop processing if fetching the block was not succesful, there is no block data,
       // or if the block does not have enough transactions
-      if (!block.success || block.data.transactions.length < minTransactionCount) return acc;
+      if (
+        !block.success ||
+        !block.data ||
```
Siegrift: This should work as well, right?

```suggestion
        !block?.data?.transactions ||
```

Siegrift (approved):

src/gas-oracle.ts
vponline: Yep, but that syntax doesn't seem to have been used much elsewhere so I wrote it open.
 
Would you prefer this change?

Siegrift: I think it's nicer, but leaving it up to you.

aquarat (commented):
I think there's a bigger issue here and that is that a bug in this component caused the Airseeker execution to end.

I think it might be a good idea to add an error check/try-catch around:
in https://github.com/API3-CTT/airseeker/blob/46f87c8b02b5207277cd2bc91a6a73b04f353699/src/update-beacons.ts#L185
with a fallback to conventional pricing.

We should expect EVM RPC endpoints to give us unexpected data because (1) good practice to expect the worst and (2) we're operating across under-served chains with new implementations.

Also, separately, we may want to add `import 'source-map-support/register';` to this application to make debugging in future easier. Size won't be an issue because we're running it ourselves.

(I didn't realise Github could quote lines inline like this^, cool!)

vponline:
> I think it might be a good idea to add an error check/try-catch around:

Thanks for checking and good catch @aquarat !

Could this be wrapped in a `go`-function with 0 retries? I'm asking because I just finished a task to replace all try-catches in Airnode with promise-utils `go`?  I'm not sure if there are any plans to replace try-catches here, though.

aquarat:
> Could this be wrapped in a `go`-function with 0 retries? I'm asking because I just finished a task to replace all try-catches in Airnode with promise-utils `go`?

I really like the go utils approach to error handling... it's very C/Go like and satisfies @Siegrift 's suggestion of keeping "unhappy paths" indented and the happy path as left as possible. I say great idea :)

Siegrift:
> Also, separately, we may want to add import 'source-map-support/register'; to this application to make debugging in future easier. Size won't be an issue because we're running it ourselves.

I think this is something we should consider for all Airnode, Airkeeper and Airseeker in the future.

vponline:
Ok I've done a bit of a refactor, here are the additions:

1. added `source-map-support/register`
2. extracted fallback gas price fetching into a new function: `getFallbackGasPrice`
3. added `go`-wrapper and fallback gas price logic to `update-beacons.ts` if `getOracleGasPrice` fails
4. added tests for the above

Please let me know what you think.

src/fetch-beacon-data.ts
```diff
@@ -1,3 +1,4 @@
+import 'source-map-support/register';
```
Siegrift: I believe we only need this in entrypoint of an application. Why is imported in multiple times?

vponline: Oh, importing only in `main.ts`?

Siegrift: At least that's how I understand their README. Be sure to test that it works.

vponline: Yep, just tested and indeed works.

aquarat: Well I just learnt something :sweat_smile:

vponline: I moved it to `index.ts` to be at the actual entrypoint file :smile:

src/update-beacons.ts
```diff
@@ -181,8 +182,32 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorBeaco
...
+        { retries: 0 }
+      );
+      if (!goFallbackGasPrice.success) {
+        gasPrice = parsePriorityFee(config.chains[chainId].options.gasOracle.fallbackGasPrice);
+      } else {
+        gasPrice = goFallbackGasPrice.data;
+      }
+    } else {
+      gasPrice = goGasPrice.data;
+    }
```
aquarat: This logic should probably have been split out into a separate function: this is about getting the gas price rather than creating the update transaction (the concern of the parent function's scope). It'd also get rid of this `let` and would be more memory efficient (it would give the GC tighter scopes).

```
// pseudo-code
const getGasPrice = async () => { 
const oracleGasPrice = await getOracleGasPrice();
if (oracleGasPrice) {
 return oracleGasPrice;
}

  return getFallbackGasPrice();
};

// ...
const gasPrice = await getGasPrice();

```

Apart from that super neat work :)

aquarat (commented):
See comment :smile:

src/update-beacons.ts
vponline: Thanks for the example, this is even neater :smile: 

I have extracted into a `getGasPrice` function and added tests for the new function. Now we should have covered the happy path and paths where both `getGasOracle` and `getFallbackGasPrice` might unexpectedly throw.

aquarat (approved):
This is really neat :1st_place_medal: (and I think it'll be both efficient and robust).
Thanks again :smile:

PR #171, opened 2022-07-21
Upgrade airseeker dependencies

### LINKED ISSUE #160: Update Airnode (and other) dependencies
There are stale PRs from dependabot updating Airnode (and other) dependencies that are failing. They require some manual changes.

acenolaza:
Closes  #160

I had to copy over the zod schemas from Airnode because the [zod library used there](https://github.com/api3dao/airnode/blob/v0.7.2/packages/airnode-validator/package.json#L33) is way to old and it is no longer compatible with the latest version of zod.

Another issue was that [chainOptionsSchema](https://github.com/api3dao/airnode/blob/v0.7.2/packages/airnode-validator/src/config/config.ts#L85) was changed from a `z.object` to a `z.discrimitantedUnion` [here](https://github.com/api3dao/airnode/pull/1103) and this means that methods like `merge` or `extend` are no longer available. Even though I could just have used `z.interection` to add the `gasOracle` schema the validation still failed because each option of the discriminated union was calling `strict()`.

Bottom line is that Airseeker no longer depends on zod schemas from Airnode and I believe that's a good thing :)


package.json
```diff
@@ -47,11 +47,11 @@
...
-    "@api3/airnode-abi": "^0.6.2",
-    "@api3/airnode-node": "^0.6.2",
-    "@api3/airnode-protocol-v1": "^0.5.0",
-    "@api3/airnode-utilities": "^0.6.2",
-    "@api3/airnode-validator": "^0.6.2",
+    "@api3/airnode-abi": "^0.7.2",
+    "@api3/airnode-node": "^0.7.2",
+    "@api3/airnode-protocol-v1": "^0.6.0",
+    "@api3/airnode-utilities": "^0.7.2",
+    "@api3/airnode-validator": "^0.7.2",
```
Siegrift: I guess we no longer need Airnode validator in dependencies.

src/gas-oracle.test.ts
```diff
@@ -59,7 +59,7 @@ const config: Config = {
         },
       },
       options: {
-        txType: 'legacy',
+        txType: 'eip1559',
```
Siegrift, line 62: Is this correct? Previously the value was `legacy`.

Siegrift (commented):
> I had to copy over the zod schemas from Airnode because the [zod library used there](https://github.com/api3dao/airnode/blob/v0.7.2/packages/airnode-validator/package.json#L33) is way to old and it is no longer compatible with the latest version of zod.

I also faced a similar issue when updating Airnode dependencies. See: https://github.com/api3dao/airnode/pull/1325#issuecomment-1191772400

> Bottom line is that Airseeker no longer depends on zod schemas from Airnode and I believe that's a good thing :)

I would not say so. Similar argument could be said about other airnode-xyz dependency, but reusing that code is simply convenient. But I support your decision by copying over the necessary validations because of the "zod" issues.

package.json
acenolaza: Good catch!

src/gas-oracle.test.ts
acenolaza: It used to be `legacy` but it had the properties of the `eip1559` so I assume that either was not being used in those tests or it was using the properties of `eip1559` but had the wrong type set 🤷🏻

acenolaza:
> > I had to copy over the zod schemas from Airnode because the [zod library used there](https://github.com/api3dao/airnode/blob/v0.7.2/packages/airnode-validator/package.json#L33) is way to old and it is no longer compatible with the latest version of zod.
> 
> I also faced a similar issue when updating Airnode dependencies. See: [api3dao/airnode#1325 (comment)](https://github.com/api3dao/airnode/pull/1325#issuecomment-1191772400)
> 
> > Bottom line is that Airseeker no longer depends on zod schemas from Airnode and I believe that's a good thing :)
> 
> I would not say so. Similar argument could be said about other airnode-xyz dependency, but reusing that code is simply convenient. But I support your decision by copying over the necessary validations because of the "zod" issues.

Right 😁 I meant in this particular case

Siegrift (approved):
👍 LGTM

Issue #167, opened 2022-07-17
Refactor `update-beacon-sets.ts` & `update-beacons.ts` to reuse common parts


amarthadan:
Once https://github.com/API3-CTT/airseeker/issues/166 is done.

PR #182, opened 2022-08-02
Implemet beacon set updates #166

### LINKED ISSUE #166: Implemet beacon set updates
I'd say that the implementation will be different enough to be done separately from `update-beacons.ts`. I would prefer starting with a separate implementation in `update-beacon-sets.ts` and then refactor both and reuse common parts.

The flow of beacon set update:
* retrieve values for each beacon within the set from the cache (common memory)
* IF there's no value for a given beacon, fetch it from the chain
* IF the value is not available on the chain skip the update
* fetch beacon set value & timestamp from the chain
* calculate beacon set timestamp from beacon timestamps (https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/dapis/DapiServer.sol#L443)
* IF new timestamp is older than the on-chain one skip the update
* IF the last update is older than now + heartbeat interval force update
* IF the deviation threshold is reached do the update, skip otherwise

Use the [`readDataFeedWithId`](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/dapis/DapiServer.sol#L708) function to rad the data (both beacon and beacon sets) and [`updateBeaconSetWithSignedData`](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/dapis/DapiServer.sol#L569) function to update beacon sets.

acenolaza:
Closes #166 

I tried using the same flow and similar logic used for updating beacon values. This should come handy when working on #167 

I've left a couple of TODOs with some doubts that I have.

This is a draft PR basically because I haven't been able to execute the run-locally.ts script so I need to find a proper way to test these changes locally and also add an e2e test case.


acenolaza:
While trying to fix the e2e tests, I've realized that it might not be a good idea to run the beacon and beacon-set updates in parallel because we might get nonce errors when using the same sponsor wallet. They seem to be needed to run one after the other.

Siegrift:
> They seem to be needed to run one after the other.

But how to achieve this? Can't we instead force different sponsor wallets for beacons and beacon sets?

acenolaza:
> > They seem to be needed to run one after the other.
> 
> But how to achieve this? Can't we instead force different sponsor wallets for beacons and beacon sets?

I'm refactoring the airseeker.json config file where I'm now grouping `beacons` and `beaconSets` triggers by `sponsorWallet`. 

Managing sponsor wallets and keeping the topped up is not a trivial task so I'm not sure if forcing them to use separate sponsor wallets is a good option 🤷🏻

Siegrift:
> I'm refactoring the airseeker.json config file where I'm now grouping beacons and beaconSets triggers by sponsorWallet.

So they will all have the same `updateInterval ` value, right?

acenolaza:
> > I'm refactoring the airseeker.json config file where I'm now grouping beacons and beaconSets triggers by sponsorWallet.
> 
> So they will all have the same `updateInterval ` value, right?

Exactly, that's the only "downside" that I've found so far

Siegrift:
> that's the only "downside" that I've found so far

I don't see a big problem with it. Btw. are you working on https://api3workspace.slack.com/archives/C037NLX1R8T/p1658137012920379?thread_ts=1658072085.473819&cid=C037NLX1R8T as well?

acenolaza:
> > that's the only "downside" that I've found so far
> 
> I don't see a big problem with it. Btw. are you working on https://api3workspace.slack.com/archives/C037NLX1R8T/p1658137012920379?thread_ts=1658072085.473819&cid=C037NLX1R8T as well?

Yes and I'm about to push a commit with this new refactor (in a separate branch for just in case)
btw, thanks for taking the time to review this PR 🙏🏻

src/fetch-beacon-data.ts
```diff
@@ -17,14 +17,12 @@ export const initiateFetchingBeaconData = async () => {
...
-        return beaconUpdate.beacons.flatMap((b) => b.beaconId);
-      });
-    }),
-    ...Object.values(config.triggers.beaconSetUpdates).flatMap((beaconSetUpdatesPerSponsor) => {
-      return Object.values(beaconSetUpdatesPerSponsor).flatMap((beaconSetUpdate) => {
-        return beaconSetUpdate.beaconSets.flatMap((b) => config.beaconSets[b.beaconSetId]);
+    ...Object.values(config.triggers.dataFeedUpdates).flatMap((dataFeedUpdatesPerSponsor) => {
+      return Object.values(dataFeedUpdatesPerSponsor).flatMap((dataFeedUpdate) => {
+        return [
+          ...dataFeedUpdate.beacons.flatMap((b) => b.beaconId),
```
amarthadan: I think this one doesn't need to be `flatMap` but just `map`, the returning value is a string.

src/update-data-feeds.ts
```diff
@@ -0,0 +1,459 @@
...
+  const { config, beaconValues } = getState();
+  const { provider, sponsorAddress, beacons } = providerSponsorBeacons;
+  const { rpcProvider, chainId, providerName } = provider;
+  const logOptionsSponsor = {
+    meta: { chainId, providerName },
+    additional: { Sponsor: shortenAddress(sponsorAddress) },
+  };
+  logger.debug(`Processing beacon updates`, logOptionsSponsor);
+
+  const startTime = Date.now();
```
amarthadan: Having the `startTime` initialized within `updateBeacons` and `updateBeaconSets` separately effectively doubles the time the processing can take. The start time should be passed to both from the `updateDataFeedsInLoop` function.

src/update-data-feeds.ts
```diff
@@ -0,0 +1,459 @@
...
+        additional: {
+          ...logOptionsBeaconSetId.additional,
+          'Beacon-ID': beaconId,
+        },
+      };
+
+      const beaconResponse: SignedData = beaconValues[beaconId];
+
+      // Check whether we have a value for given beacon
+      if (isNil(beaconResponse)) {
```
amarthadan: Do we need to use `isNil`? Isn't `if (!beaconValues[beaconId])` enough?

src/update-data-feeds.ts
```diff
@@ -0,0 +1,459 @@
...
+        // IF there's no value for a given beacon, fetch it from the chain
+        const beaconValueOnChain = await readDataFeedWithId(
+          voidSigner,
+          contract,
+          beaconId,
+          prepareGoOptions(startTime, totalTimeout),
+          logOptionsBeaconId
+        );
+        // IF the value is not available on the chain skip the update
+        if (isNil(beaconValueOnChain)) {
```
amarthadan: Same here.

src/update-data-feeds.ts
```diff
@@ -0,0 +1,459 @@
...
+        const message = `New beacon value is out of type range.`;
+        logger.warn(message, logOptionsBeaconId);
+        throw new Error(message);
+      }
+
+      return { beaconId, ...beaconResponse, value: decodedValue, ...config.beacons[beaconId] };
+    });
+    const beaconSetBeaconValuesResults = await Promise.allSettled(beaconSetBeaconValuesPromises);
+    if (beaconSetBeaconValuesResults.some((data) => data.status === 'rejected')) {
+      logger.warn('There was an error fetching beacon data for beacon set. Skipping.');
```
amarthadan: Don't you need to pass the log options so we know what beacon set is this in regard to?

src/update-data-feeds.ts
```diff
@@ -0,0 +1,459 @@
...
+    // fetch beacon set value & timestamp from the chain
+    const beaconSetValueOnChain = await readDataFeedWithId(
+      voidSigner,
+      contract,
+      beaconSetUpdate.beaconSetId,
+      prepareGoOptions(startTime, totalTimeout),
+      logOptionsBeaconSetId
+    );
+    // TODO: what if the beaconSet is not yet initialized?
+    // Should we continue with next beaconSet? See https://github.com/API3-CTT/airseeker/blob/166-update-beacon-sets-implementation/src/update-beacons.ts#L151
```
amarthadan: Is the beacon set not being initialized a problem? Shouldn't we, in that case, update it to initialize it? @bbenligiray

src/update-data-feeds.ts
```diff
@@ -0,0 +1,459 @@
...
+      // IF the deviation threshold is reached do the update, skip otherwise
+      const median = (arr: ethers.BigNumber[]) => {
+        const mid = Math.floor(arr.length / 2);
+        const nums = [...arr].sort((a, b) => {
+          if (a.lt(b)) return -1;
+          else if (a.gt(b)) return 1;
+          else return 0;
+        });
+        return arr.length % 2 !== 0 ? nums[mid] : nums[mid - 1].add(nums[mid]).div(2);
+      };
```
amarthadan: You can move the function outside but I presume you're planning to do that as part of https://github.com/api3dao/airseeker/issues/167

src/update-data-feeds.ts
```diff
@@ -0,0 +1,459 @@
...
+      // Check beacon set condition
+      // TODO: should we use median like in DapiServer or mean?
```
amarthadan: I think we should use the same as the contract is using. @bbenligiray

src/update-data-feeds.test.ts
```diff
@@ -165,85 +180,90 @@ const config: Config = {
...
+type ComparableProviderSponsorDataFeeds = {
   provider: string;
   sponsorAddress: string;
   updateInterval: number;
   beacons: BeaconUpdate[];
+  beaconSets: BeaconSetUpdate[];
 };
 
-const cpsbg: ComparableProviderSponsorBeacons[] = [
+const cpsbg: ComparableProviderSponsorDataFeeds[] = [
```
amarthadan: Can you change the variable name accordingly?

amarthadan (changes requested):
Looks good :+1: Just some minor issues.

src/update-data-feeds.ts
bbenligiray: Exact same implementation as the contract

src/update-data-feeds.ts
acenolaza: Right, I wanted to first confirm this is the right function to call and I also kept is close to where it is being used to make it easier to the reviewer. For now I'll just move it outside this function but I'll keep it in the same file until #167

src/update-data-feeds.ts
acenolaza: I think it shouldn't be an issue and we should continue to initialize it but this validation [here](https://github.com/api3dao/airseeker/blob/166-update-beacon-sets-implementation/src/update-data-feeds.ts#L164-L166) got me confused

src/update-data-feeds.ts
acenolaza: Yes,  we do. Good catch!

src/update-data-feeds.ts
acenolaza: I'm using a variable because we are accessing that element in the array a couple of times down below and yes, using `!` instead of `isNil` should have the same effect BUT 🤭 I have trusting issues when it comes to javascript evaluating truthiness for things like empty objects `{}` being truthy, etc but I can stop using `isNil` if there are performance issues or something.

src/update-data-feeds.ts
acenolaza: Good point 🤔 I wonder if this would be better handled in #167. I'll probably just apply your suggestion for the time being tho

src/update-data-feeds.ts
amarthadan: Not a performance issue really. It just threw me off seeing using `isNil` and `!` for essentially the same check (one in `updateBeacons` and one in `updateBeaconSets`). I would just like to keep it consistent.

amarthadan (approved):

src/update-data-feeds.ts
acenolaza: @bbenligiray any comments around this TODO?

bbenligiray: I'm not sure what validation you're referring to but I don't see any reason to not initialize it

acenolaza: I was referring to the check for an onchain value in order to update a beacon [here](https://github.com/api3dao/airseeker/blob/main/src/update-beacons.ts#L151).

bbenligiray: There are three things that may happen
- Successfully read an initialized data feed with a value and non-zero timestamp (an return the value)
- Successfully read an uninitialized data feed with zero value and zero timestamp (and return the value)
- Fail reading the data feed

The problem here is that when `readOnChainBeaconData()` returns zero, you won't know if the data feed is initialized or not. Normally it should initialize the data feed no matter what the deviation is. Note that `readOnChainBeaconData()` returning `null` doesn't mean that it's not initialized. Also that function should probably be called `readOnChainDataFeed()`.

acenolaza: Right, the issue is not around the data feed being initialized or not but instead checking if Airseeker is able to read the data feed onchain value or not.

Meaning that if [beaconSetValueOnChain](https://github.com/api3dao/airseeker/blob/202434c52b341b2842bc6a7ecdeb48596090dca7/src/update-data-feeds.ts#L372) is `null` then there was probably an unexpected issue that prevented Airseeker from reading the onchain value (provider error, msg.sender not being whitelisted, etc). 

So the questions is: should we proceed with the update if Airseeker failed to read the onchain value? or just continue with the next `beaconSet` (like when updating beacons)?

bbenligiray: Ah sorry, I thought this was about initialization. If the returned value is `null` we should skip it.

acenolaza: Cool, thanks 🙏🏻  and no need to apologize since I introduced the confusion by talking about the value not being initialized in the TODO comment.

PR #183, opened 2022-08-03
Switch to gas oracle from airnode-utilities

### LINKED ISSUE #220: Implement strategies for gas price oracle
For context: 

https://api3workspace.slack.com/archives/C037NLX1R8T/p1653813343064269

Strategies will be applied in the order they are available in the configuration file.

We want to support 3 strategies, latestBlockPercentileGasPrice, providerRecommendedGasPrice, and constantGasPrice all of which are already implemented in gas price oracle. What is needed is to move them into independent building blocks so they can be used in any order.

The fee type will be deduced based on the gas price strategy.

https://github.com/API3-CTT/airseeker/pull/183

vponline:
Switch to using the gas oracle from `airnode-utilities`, including strategies.

Draft until Airnode 0.8 is published.


vponline:
I have rebased this on top of #238 and removed the Airseeker Gas Oracle. I also had to update the airnode-logger types/input formats and removed unnecessary tests.

@acenolaza I ran the E2E tests locally in a loop for a while and didn't get any errors but please also give it a try and let me know if you are still able to reproduce it.

src/update-data-feeds.ts
```diff
@@ -255,8 +259,13 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
       logger.info(`Deviation threshold reached. Updating.`, logOptionsBeaconId);
     }
 
-    // Get gas price from oracle
-    const gasPrice = await getGasPrice(provider, config.chains[chainId].options.gasOracle);
+    // Get the latest gas price
+    const [logs, gasTarget] = await getGasPrice(provider.rpcProvider, config.chains[chainId].options);
+    logs.forEach((log) =>
+      log.level === 'ERROR'
+        ? logger.error(log.message + log.error ? `Error: ${log.error?.message}` : '')
```
acenolaza: Shouldn't this be something like:

```suggestion
        ? logger.error(log.error ? `Error: ${log.error.message}` : 'log.message')
```

acenolaza (approved):
This PR looks really neat 🥇

src/update-data-feeds.ts
acenolaza: or maybe even: 

```suggestion
        ? logger.error(log.error && log.error.message ? `Error: ${log.error.message}` : 'log.message')
```

vponline: Good catch! The returned logs from airnode-utilities will actually already include the error message if there is one, e.g: [here](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts#L279).

So perhaps it could just be 
```
logs.forEach((log) => (log.level === 'ERROR' ? logger.error(log.message) : logger.info(log.message)));
```

PR #184, opened 2022-08-04
Refactor beaconSet updates

acenolaza:
This is the refactor mentioned [here](https://github.com/API3-CTT/airseeker/pull/182#issuecomment-1205298753) where I've made changes in order to make sure all txs using the same sponsor wallet are run serially to avoid nonce issues.


acenolaza:
I'm merging this to the parent branch since we seemed to agree that this is a good enough approach to fix the nonce issue when sending multiple txs with the same sponsorWallet. https://api3workspace.slack.com/archives/C037NLX1R8T/p1660048084017919

PR #205, opened 2022-08-19
Remove unnecesary parts

### LINKED ISSUE #162: Remove old/unnecesary parts
Since we're not running Airseeker as a long-running service, there are parts that can be removed:
* Terraform recipes
* pm2 setup
* Docker containerization
* some helper scripts

We can (and probably should) remove these to keep the repository clean and less confusing at first sight.
@aquarat @bbenligiray What do you think about this? I'd say we should remove what we can, we can always dig it back up from the git history if needed :man_shrugging:

amarthadan:
Close #162

I'm pretty sure all this can be removed but whoever will be reviewing this, please, read it carefully :smile:

I also fixed the tests not passing when the project is built before running them (we don't do that in the CI, but it's quite common for a local setup) in the first commit.


README.md
```diff
@@ -47,62 +47,63 @@ secrets file.
...
-[generated logrotate configuration from PM2](https://pm2.keymetrics.io/docs/usage/log-management/#setting-up-a-native-logrotate).
+The deployed Airseeker function won't immediately start (due to the scheduler).
 
-### Build
+There are two main strategies for dealing with this:
 
-```sh
-yarn docker:build
-```
+1. Deploy Airseeker using and start the new Airseeker immediately using the invoke commands below.
```
Siegrift:
```suggestion
1. Deploy Airseeker using WHAT and start the new Airseeker immediately using the invoke commands below.
```

Siegrift: same below

README.md
```diff
@@ -47,62 +47,63 @@ secrets file.
...
+2. Deploy Airseeker using and wait up to 14 minutes for the deployed Airseeker to start.
```
Siegrift: Mhhh, why 14?

test/tsconfig.json
```diff
@@ -1,9 +1,11 @@
 {
-  "extends": "../tsconfig.base.json",
+  "extends": "../tsconfig.json",
```
Siegrift, line 2: Nit: We might want to get rid of the TS config base (inline it similarly to how its done in Airnode).

Leaving that up to you.

README.md
```diff
@@ -47,62 +47,63 @@ secrets file.
...
+```shell
+yarn sls remove
 ```
 
-### Steps to remove from AWS
+**Update Airseeker**
 
-Then to destroy the deployment you can run the following command:
+If an Airseeker is already deployed, deploy the new Airseeker using a different service name in `serverless.yaml`, so
+that both Airseekers can run concurrently. Wait for the new Airseeker to start. Once it does, ensure that you're
```
Siegrift, line 98: How can you tell if the second one works correctly? 

The onchain updates could be done by the first (older) Airseeker. You can check on AWS whether the new Airseeker is running fine, though.

Siegrift (commented):
Some minor comments, but 👍 LGTM.
I'll try to test this (and only give approve after I've seen it work).

.gitignore
```diff
@@ -3,8 +3,12 @@
 # Package directories & files
 /node_modules
 *.tgz
-# Ignore dist folder in every (sub)folder
+# Ignore dist build folder in every (sub)folder
```
acenolaza:
```suggestion
# Ignore dist and build folder in every (sub)folder
```

README.md
acenolaza: Probably because of [this](https://github.com/api3dao/airseeker/blob/59dac797f8148c56945a509fd8514be213f93a9b/serverless.aws.yml#L27)

README.md
acenolaza:
```suggestion
1. Deploy Airseeker and start the new Airseeker immediately using the invoke commands below.
```

README.md
acenolaza:
```suggestion
2. Deploy Airseeker and wait up to 14 minutes for the deployed Airseeker to start.
```

README.md
```diff
@@ -47,62 +47,63 @@ secrets file.
...
-# Starting Airseeker
-yarn pm2:start
-# PM2 status
-yarn pm2:status
-# Logs
-yarn pm2:logs
-# Stopping Airseeker
-yarn pm2:stop
-```
+Export AWS credentials in your terminal or ensure your local aws installation is logged in.
```
acenolaza, line 48: Should we point to some aws or api3 docs where we mention which vars we are talking about? (AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY)

README.md
```diff
@@ -47,62 +47,63 @@ secrets file.
...
-```
+1. Deploy Airseeker using and start the new Airseeker immediately using the invoke commands below.
+2. Deploy Airseeker using and wait up to 14 minutes for the deployed Airseeker to start.
 
-Resulting image is named `api3/airseeker`.
+**Deploy Airseeker:**
 
-## Deploy
+```shell
+yarn sls deploy --region us-east-1 --config serverless.aws.yml
```
acenolaza, line 62: Is the `--region` arg needed here in order to override [this](https://github.com/api3dao/airseeker/blob/59dac797f8148c56945a509fd8514be213f93a9b/serverless.aws.yml#L11) value?

README.md
```diff
@@ -47,62 +47,63 @@ secrets file.
...
 
-To run deploy Airseeker then you will need to cd into the `terraform` directory and run the following commands:
+**Remove Airseeker:**
 
-```sh
-terraform init
-terraform apply -var 'app_environment=dev'
-Alternatively you can provide the `aws_region`, `app_environment`, `app_docker_image`, etc as arguments to the `terraform apply` command.
+```shell
+yarn sls remove
```
acenolaza: Don't you need to also give it the `--config` arg value?

README.md
```diff
@@ -47,62 +47,63 @@ secrets file.
...
 ```
 
-### Steps to remove from AWS
+**Update Airseeker**
 
-Then to destroy the deployment you can run the following command:
+If an Airseeker is already deployed, deploy the new Airseeker using a different service name in `serverless.yaml`, so
```
acenolaza, line 97: I wonder if one can do this by just using a different `stage` value in serverless.yaml 🤔

acenolaza:
[This](https://github.com/api3dao/airseeker/blob/remove-unnecesary-parts/scripts/run-locally.ts#L0-L1) script might also not make sense now that the idea is to use serverless invoke to run Airseeker locally, right?

README.md
amarthadan: Yes.

README.md
amarthadan: I was wondering about that as well and according to the [documentation](https://www.serverless.com/framework/docs/providers/aws/cli-reference/remove), there's no `--config` option for the `remove` command. But I've just double-checked it in the actual CLI and the option is available. I'll add it.

amarthadan:
> [This](https://github.com/api3dao/airseeker/blob/remove-unnecesary-parts/scripts/run-locally.ts#L0-L1) script might also not make sense now that the idea is to use serverless invoke to run Airseeker locally, right?

`serverless invoke` will invoke the deployed Lambda function, it doesn't run the code locally.

README.md
amarthadan: I don't think we actually have a good section about this in the docs. There are mentions about the `aws.env` file (https://docs.api3.org/airnode/v0.7/reference/examples/aws-env.html, https://docs.api3.org/airnode/v0.7/grp-providers/tutorial/quick-deploy-aws/#aws-env) but it feels like it would be more confusing than helpful.

acenolaza:
> > [This](https://github.com/api3dao/airseeker/blob/remove-unnecesary-parts/scripts/run-locally.ts#L0-L1) script might also not make sense now that the idea is to use serverless invoke to run Airseeker locally, right?
> 
> `serverless invoke` will invoke the deployed Lambda function, it doesn't run the code locally.

I'm a bit confused now 🤭 wouldn't this work for running it locally? `yarn sls invoke local --config serverless.aws.yml -f airseeker`

amarthadan:
> > > [This](https://github.com/api3dao/airseeker/blob/remove-unnecesary-parts/scripts/run-locally.ts#L0-L1) script might also not make sense now that the idea is to use serverless invoke to run Airseeker locally, right?
> > 
> > 
> > `serverless invoke` will invoke the deployed Lambda function, it doesn't run the code locally.
> 
> I'm a bit confused now hand_over_mouth wouldn't this work for running it locally? `yarn sls invoke local --config serverless.aws.yml -f airseeker`

Oh, I haven't realized there's a `local` option, sorry. Yes, in that case the script and the `start` target don't need to be there. I'll remove it.

acenolaza (approved):
LGTM 👍🏻

Siegrift (changes requested):
Running locally via `yarn sls invoke local --config serverless.aws.yml -f airseeker` fails for me with:
```
$ /Users/emanueltesar/vacuumlabs/api3/airseeker/node_modules/.bin/sls invoke local --config serverless.aws.yml -f airseeker
✖ Error: Cannot find module '/Users/emanueltesar/vacuumlabs/api3/airseeker/dist/serverless'
...
```

the problem is that the `/dist` folder has the sources wrapped in `src` folder. Changing serverless config path [here](https://github.com/api3dao/airseeker/blob/remove-unnecesary-parts/serverless.aws.yml#L21) helps, but we get another error:
```
Error:
Error: Failed to parse config file. Error: ENOENT: no such file or directory, open '/Users/emanueltesar/vacuumlabs/api3/airseeker/dist/config/airseeker.json'
...
```

since the config dir is not included in dist.

amarthadan:
> Running locally via `yarn sls invoke local --config serverless.aws.yml -f airseeker` fails for me with:
> 
> ```
> $ /Users/emanueltesar/vacuumlabs/api3/airseeker/node_modules/.bin/sls invoke local --config serverless.aws.yml -f airseeker
> ✖ Error: Cannot find module '/Users/emanueltesar/vacuumlabs/api3/airseeker/dist/serverless'
> ...
> ```
> 
> the problem is that the `/dist` folder has the sources wrapped in `src` folder. Changing serverless config path [here](https://github.com/api3dao/airseeker/blob/remove-unnecesary-parts/serverless.aws.yml#L21) helps, but we get another error:
> 
> ```
> Error:
> Error: Failed to parse config file. Error: ENOENT: no such file or directory, open '/Users/emanueltesar/vacuumlabs/api3/airseeker/dist/config/airseeker.json'
> ...
> ```
> 
> since the config dir is not included in dist.

Good catch! There were multiple issues actually. Should be fixed now.

Siegrift (approved):
Works for me locally.

acenolaza:
Something that I find a bit strange is that after running `yarn build` we get a `/build` and a `/dist` folders in the root. Shouldn't we only have one or the other? (I see that `src-tsconfig.tsbuildinfo` file is being put inside `/build` on purpose)

amarthadan:
> Something that I find a bit strange is that after running `yarn build` we get a `/build` and a `/dist` folders in the root. Shouldn't we only have one or the other? (I see that `src-tsconfig.tsbuildinfo` file is being put inside `/build` on purpose)

@Siegrift will know better, so, correct me if I'm wrong but I think `/build` is for building all the source files so they are typechecked and `/dist` contains only what we would package in the NPM package.

Siegrift:
> @Siegrift will know better, so, correct me if I'm wrong but I think /build is for building all the source files so they are typechecked and /dist contains only what we would package in the NPM package.

Exactly.

Issue #214, opened 2022-08-23
Change E2E tests to use Airseeker as a service


umersin61:
Currently, airseeker.feature.ts is importing the main() function to run the Airseeker for each test.

This change should be that Airseeker is started as a pm2 service and there is a single sleep() period for each test to wait for one Airseeker cycle to finish after the contracts are re-deployed.  (We can experiment with how long the sleep should be, probably 10-15 seconds)

amarthadan:
I don't really think this will bring much anymore. See https://github.com/api3dao/airseeker/issues/161

acenolaza:
> I don't really think this will bring much anymore. See https://github.com/api3dao/airseeker/issues/161

I agree. We should close this issue @umersin61

PR #222, opened 2022-08-25
Scripts and instructions for running Airseeker locally

### LINKED ISSUE #170: Implement environment setup for testing
Even though we have E2E tests we from time to time need to test new Airseeker features so we need to set up an environment (prepare blockchain, run Airnode(s), etc.) to do it. It would be nice to have this at least semi-automated/scripted.

acenolaza:
Closes #170 

This a very basic and simple setup that will let devs run Airseeker locally by following a few steps in a matter of seconds. This will be useful when trying to test new implementations or bug fixes. 

When Airnode v0.8 package is published we can make this setup more customizable and improve the scripts to maybe allow user inputs to help create a dynamic config file. The idea is to make a better use of the local signed data gateway that was implemented in that version of Airnode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ re-remove the CloudFormation stack.
 2. Next step is to run this script:
 
    ```shell
-   yarn run dev:create-local-config
+   yarn run dev:setup-local-node
    ```
 
    This will deploy the DapiServer contract to the local ethereum node and also send funds to a test Airseeker sponsor
--- a/README.md
+++ b/README.md
@@ -18,9 +18,7 @@ yarn build
 
 You need to create a configuration file `config/airseeker.json`. Take a look at `config/airseeker.example.json` for an
 example configuration file. You can use string interpolation (with `${VAR}` syntax) for providing secrets. Secrets are
-read from the environment variables. When running locally, either just with `yarn start` or via process manager, secrets
-are automatically loaded from `config/secrets.env` file. Take a look at `config/secrets.example.env` for an example
-secrets file.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -107,3 +99,45 @@ check the resources tab of the stack in question to see errors. Manually remove
...
+   yarn run dev:testing-services:start
+   ```
+
+   A local ethereum node will be started along with a node.js express server that will return signed data (basically a
+   substitute to the Airnode's signed data gateway).
+
+2. Next step is to run this script:
+
+   ```shell
+   yarn run dev:create-local-config
```
amarthadan:
```suggestion
   yarn run dev:setup-local-node
```

amarthadan (approved):
I very much like this :+1:

This will be used just for "manual testing" when we want to try Airseeker locally, right? The scripts from `test/setup` are still needed for tests. I presume this and those scripts will overlap in a lot of code. We should probably refactor that later on.

acenolaza:
> I very much like this +1
> 
> This will be used just for "manual testing" when we want to try Airseeker locally, right?

Yes. The intent is that these scripts are only used for that purpose.

> The scripts from `test/setup` are still needed for tests. I presume this and those scripts will overlap in a lot of code. We should probably refactor that later on.

Correct and currently they are almost identical in content but when we start using the v0.8 http signed gateway **maybe** the config and secrets generation will be different. Still, you might be right and we could extract the DapiServer deployment (and maybe the sponsor wallet funding too) and use it both places since that's unlikely going to change. This  issue https://github.com/api3dao/airseeker/issues/214 might be a good opportunity to think about this refactor I think.

PR #236, opened 2022-09-13
Imports airnode-protocol to access protocol constants

### LINKED ISSUE #217: Use protocol constants from Airnode inside Airseeker
We have added protocol indices as part of v0.8 Airnode release. We should use these constants inside Airseeker.

https://github.com/api3dao/airnode/issues/1195

acenolaza:
Closes #217 
Similar to [#135](https://github.com/api3dao/operations/pull/135) in operations

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
   "dependencies": {
     "@api3/airnode-abi": "^0.7.2",
     "@api3/airnode-node": "^0.7.2",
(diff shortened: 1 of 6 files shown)


amarthadan (approved):
Didn't know we had these in `airnode-protocol` :sweat_smile: Nice :slightly_smiling_face:

PR #238, opened 2022-09-14
Sets the block.timestamp to Date.now() after resetting hardhat network

### LINKED ISSUE #224: Investigate & fix flaky E2E tests
When I run E2E tests multiple times they sometimes pass and sometimes fail. This needs to be investigated and fixed.

acenolaza:
Closes #224


acenolaza:
Ended up making this PR a draft because e2e are unexpectedly failing and I need to determine if it is related to these changes.

I do see a bunch of errors like these:

```
    console.log
      [2022-09-14 21:40:26.262] WARN Unable to get transaction count. Error: Error: Full timeout exceeded

    console.log
      [2022-09-14 21:40:26.264] WARN Unable to get block number. Error: Error: Full timeout exceeded 
```

acenolaza:
After a lot of testing and discussing this issue with @vponline, I realized that the timeouts where mostly due to the `updateInterval` in the test config were too short. New value might be too high but it seems to have fixed the E2E tests so we might be able to reduce it in the future if they take too long to finish.

vponline (approved):
LGTM :1st_place_medal:

PR #242, opened 2022-09-20
Adds workaround to support fetching signed data from http gateways wi…

acenolaza:
…th version older than v0.8.0

Required for being able to deploy a new version of Airseeker that expects v0.8.0 Airnode packages but uses older deployed Airnodes
https://github.com/api3dao/operations/issues/159


aquarat (approved):
Fantastic, thank you for doing this.
Very neatly handled @ `parsedLegacy.success`

PR #243, opened 2022-09-21
Spread `getLogOptions()` object before merge() to avoid mutating the underlying object

acenolaza:


vponline (approved):
LGTM thanks for fixing (and my apologies :sweat_smile:)

PR #251, opened 2022-09-27
Update multiple individual Beacons with a single transaction #216

### LINKED ISSUE #216: Update multiple individual Beacons with a single transaction | Airseeker
Currently the Airseeker updates each Beacon with an individual transaction. This is inefficient because the 21,000 base transaction gas cost is being paid for each update. We should instead batch these updates in a single transaction. This includes the development of a proxy Multicall contract that doesn’t revert if an individual update fails. See the thread for more information

https://api3workspace.slack.com/archives/C037NLX1R8T/p1653479838267289

acenolaza:
Closes #216

### DIFF
--- a/src/update-data-feeds.ts
+++ b/src/update-data-feeds.ts
@@ -1,16 +1,22 @@
 import { Api3ServerV1__factory as Api3ServerV1Factory } from '@api3/airnode-protocol-v1';
-import { go } from '@api3/promise-utils';
 import { getGasPrice } from '@api3/airnode-utilities';
+import { go } from '@api3/promise-utils';
 import { ethers } from 'ethers';
-import { isEmpty, isNil } from 'lodash';
+import { chunk, isEmpty, isNil } from 'lodash';
 import { calculateBeaconSetTimestamp, calculateMedian } from './calculations';
 import { checkFulfillmentDataTimestamp, checkOnchainDataFreshness, checkUpdateCondition } from './check-condition';
-import { INT224_MAX, INT224_MIN, NO_DATA_FEEDS_EXIT_CODE } from './constants';
-import { logger } from './logging';
+import {
+  DATAFEED_READ_BATCH_SIZE,
+  DATAFEED_UPDATE_BATCH_SIZE,
+  INT224_MAX,
+  INT224_MIN,
+  NO_DATA_FEEDS_EXIT_CODE,
+} from './constants';
+import { logger, LogOptionsOverride } from './logging';
 import { getState, Provider } from './state';
 import { getTransactionCount } from './transaction-count';
 import { prepareGoOptions, shortenAddress, sleep } from './utils';
-import { BeaconSetUpdate, BeaconUpdate, SignedData } from './validation';
+import { Beacon, BeaconSetUpdate, BeaconUpdate, SignedData } from './validation';
 
 type ProviderSponsorDataFeeds = {
   provider: Provider;
@@ -183,9 +189,19 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
   } = initialUpdateData;
   const { chainId } = provider;
 
-  // Process beacon updates
-  let nonce = transactionCount;
+  type BeaconUpdateData = {
+    logOptionsBeaconId: LogOptionsOverride;
+    beaconUpdate: BeaconUpdate;
+    beacon: Beacon;
+    newBeaconResponse: SignedData;
+    newBeaconValue: ethers.BigNumber;
+    dataFeedsCalldata: string;
+  };
 
+  // Process beacon read calldatas
+  let beaconUpdates: BeaconUpdateData[] = [];
+
+  // Iterate over beacons listed under triggers section in config
   for (const beacon of beacons) {
     const logOptionsBeaconId = {
       ...logOptions,
@@ -196,11 +212,10 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
       },
     };
 
-    const beaconUpdateData = { ...beacon, ...config.beacons[beacon.beaconId] };
+    logger.debug(`Processing beacon update`, logOptionsBeaconId);
 
-    logger.debug(`Updating beacon`, logOptionsBeaconId);
-    // Check whether we have a value for given beacon
-    const newBeaconResponse = beaconValues[beaconUpdateData.beaconId];
+    // Check whether we have a value from the provider API for given beacon
+    const newBeaconResponse = beaconValues[beacon.beaconId];
     if (!newBeaconResponse) {
       logger.warn(`No data available for beacon. Skipping.`, logOptionsBeaconId);
       continue;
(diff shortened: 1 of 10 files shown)


src/update-data-feeds.ts
```diff
@@ -189,9 +189,8 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
   } = initialUpdateData;
   const { chainId } = provider;
 
-  // Process beacon updates
-  let nonce = transactionCount;
-
+  // Process beacon update calldatas
+  const calldatas: string[] = [];
   for (const beacon of beacons) {
```
acenolaza: I tried replacing this for loop with Array iteration methods like `map` and `reduce` but although it works, the resulting code is harder to read and probably less performant so I just left the for loop.

acenolaza:
@bbenligiray pointed out today that `multicall` is likely going to fail if a beacon is updated right before submitting the update tx. It was suggested to add a flag to the config in order to select which method of update to use. Either multicall or previous single beacon per tx update. I'm looking at the code while thinking about the implementation details and I realized that if we are going to just keep using single beacon update per tx then it might be better to just leave this here and implement it once [`tryMulticall`](https://github.com/api3dao/airnode-protocol-v1/issues/19) is ready in DapiServer. Merging this now doesn't make much sense if it is not going to be used and current Airseeker is tested and working fine as it is.

src/update-data-feeds.ts
aquarat: Better late than never: I think this is perfectly fine 😄

bbenligiray:
Looks good to me but I'd prefer if someone else took a look because I'm not familiar with Airseeker

vponline (approved):
So this change will attempt to update 10 beacons at a time. If I remember correctly, it was mentioned on a call that a single failure in the batch won't fail the whole transaction, so the loop will continue, right? And that failed update will simply be re-tried on the next cycle (if necessary)?

acenolaza:
> So this change will attempt to update 10 beacons at a time.

I've updated the batch size to 100 [here](https://github.com/api3dao/airseeker/blob/cb7290f215b80cd62a5715b357c06eb387bf16f7/src/constants.ts#L21). 

> a single failure in the batch won't fail the whole transaction, so the loop will continue, right? And that failed update will simply be re-tried on the next cycle (if necessary)?

That is correct. There should never be a revert due to a failure in a data feed signed data udpate. With tryMulticall() we are getting back a list of successes (bool[]) and a list of returnedData (array of strings with encoded revert string or encoded data feed values). Any update that failed will be retried on the next iteration but tx won't fail due to an update reverting. Airseeker will continue processing the next batch of updates even when a specific update fails.

bbenligiray:
50,000 x 100 = 5,000,000 gas transaction sounds excessive. Read in batches of 100, write in batches of 10 imo. Also, is there a reason why this is specifically named as Beacon updates? Can this not also do Beacon set updates in batches?

acenolaza:
> Can this not also do Beacon set updates in batches?

I was wondering the same thing but I guess I should have asked about it somewhere. In the [issue description](https://github.com/api3dao/airseeker/issues/216#issue-1347554041) it talks about beacon updates only and I made the assumption that for beaconSets multicall was not required but I can also make use of multicall for beaconSet updates.

bbenligiray:
I think also including Beacon set updates is a good idea because then we would have more updates to batch together. One point of concern is that a Beacon set update costs more gas, but that is probably not an issue in a batch size of 10 (as in you could feasibly update 10 Beacon sets with 9 Beacons each in a transaction).

Whether to update this PR or create an issue about it is up to you. We don't have Beacon sets to batch-update so this will only help in the mid term. If it's easy enough, may as well not create future work, if not, merging this may be a better idea.

acenolaza:
Just a heads up, there might be further changes required for updating beacons sets: https://api3workspace.slack.com/archives/C037NLX1R8T/p1676541600224609?thread_ts=1675433240.760419&cid=C037NLX1R8T

src/update-data-feeds.ts
```diff
@@ -180,50 +195,115 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
...
-        logger.warn(`Unable to fetch current beacon value`, logOptionsBeaconId);
-        // This can happen only if we reach the total timeout so it makes no sense to continue with the rest of the beacons
-        return;
+    const { successes, returndata } = goDatafeedsTryMulticall.data;
+
+    // Process beacon update calldatas
+    let updateDataFeedWithSignedDataCalldatas: string[] = [];
+
+    for (let i = 0; i < readBatch.length; i++) {
+      const beaconReturndata = returndata[i];
```
aquarat, line 262:
```suggestion
      const beaconReturnData = returnData[i];
```
This is super minor 😅 . This would require `returndata`/`returnData` to be capitalised elsewhere.

src/update-data-feeds.ts
```diff
@@ -180,50 +195,115 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
...
+  for (const readBatch of chunk(beaconUpdates, DATAFEED_READ_BATCH_SIZE)) {
+    // Read beacon batch onchain values
+    const goDatafeedsTryMulticall = await go(
+      () => {
+        const calldatas = readBatch.map((beaconUpdateData) => beaconUpdateData.dataFeedsCalldata);
+        return contract.connect(voidSigner).callStatic.tryMulticall(calldatas);
+      },
+      {
+        ...prepareGoOptions(startTime, totalTimeout),
+        onAttemptError: (goError) =>
```
aquarat, line 247: This is pretty cool, I didn't realise this library could do this.

aquarat (approved):
I can't see anything wrong with the contents of this PR 😅 
I was a bit concerned about multicall failing due to the gas usage, but I see that that was addressed in the comments.
Well done, it's neat and you've covered all the failure scenarios with generous logging ❤️

acenolaza:
`updateDataFeeWithSignedData()` was renamed to `updateBeaconWithSignedData()` [in DapiServer](https://github.com/api3dao/airnode-protocol-v1/pull/96) will no longer work for beacon sets. @bbenligiray for visibility

src/update-data-feeds.ts
acenolaza: Yes, I know it looks weird 🤭 but I did it because I was aware of the names used in the returned variables on the [contract function](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/utils/SelfMulticall.sol#L60). Therefore I could have destructured it like this:
`const { successes, returndata: returnData } = goDatafeedsTryMulticall.data;`
but I wasn't sure if it added any value 🤷🏻‍♂️

aquarat:
So far this PR seems very efficient [(with some minor fixes applied)](https://github.com/api3dao/airseeker/pull/393).
https://mumbai.polygonscan.com/address/0x9c2914c0aeb698f618d0cd8d98f88a44eaf31b16

I'm currently running it, I'll extend it to more chains shortly and then do an actual code review. At first glance I don't see any big issues :)

aquarat (approved):
This looks great 😊, as noted this is being run against testnets currently and it's working well, it appears to be very efficient in terms of tx costs.

I would perhaps at a later stage look at breaking up the blocks of code in update-data-feeds into smaller functions because currently it's a huge block of code - but it's perfectly readable and there's no point in doing this now imo/business value.

You've done a great job generally and it's cool you merged in the rate limiting stuff without breaking anything :D 

There are naturally still some broken tests, which I'm glad I don't have to fix (or maybe I do? 🤣)

acenolaza:
> This looks great blush, as noted this is being run against testnets currently and it's working well, it appears to be very efficient in terms of tx costs.

multicall FTW! 🍾 

> I would perhaps at a later stage look at breaking up the blocks of code in update-data-feeds into smaller functions because currently it's a huge block of code - but it's perfectly readable and there's no point in doing this now imo/business value.

Completely agree

> You've done a great job generally and it's cool you merged in the rate limiting stuff without breaking anything :D

😍 

> There are naturally still some broken tests, which I'm glad I don't have to fix (or maybe I do? rofl)

Test should be fixed now 💪🏻

src/update-data-feeds.ts
```diff
@@ -312,172 +330,265 @@ export const updateBeaconSets = async (providerSponsorBeacons: ProviderSponsorDa
...
-        return { beaconId, ...beaconResponse, value: decodedValue, ...config.beacons[beaconId] };
+        beaconSetBeaconUpdateData = {
+          beaconSetBeaconValues: [...beaconSetBeaconUpdateData.beaconSetBeaconValues, { value, timestamp }],
+          updateBeaconWithSignedDataCalldatas: [
+            ...beaconSetBeaconUpdateData.updateBeaconWithSignedDataCalldatas,
+            ...(calldata ? [calldata] : []),
+          ],
+        };
+      }
+      if (shouldSkipBeaconSetUpdate || beaconSetBeaconUpdateData.updateBeaconWithSignedDataCalldatas.length === 0) {
```
acenolaza: @aquarat while fixing the e2e tests I've realized that adding this check might be a mistake 🤔 

`beaconSetBeaconUpdateData.updateBeaconWithSignedDataCalldatas.length === 0`

In the case the beaconSet beacons get updated individually then this condition might be `true` (causing a beaconSet update to be skipped) but the beaconSet might still need to be updated even if no beaconSet beacons are in need of an update. WDYT?

aquarat: Sorry, saw this this morning.
I'm struggling to think in what scenarios this might happen 😅 
I think if the underlying feeds are updated by a third party, this could happen - so the check should probably be removed. The underlying feeds being updated by a third party is definitely a feasible and potentially common scenario. I think it's also possible for this to happen if the Airseeker updates an underlying beacon separately, as part of the single beacon update flow (although we currently don't expect to operate Airseeker like this in production for managed feeds).

acenolaza: > I think it's also possible for this to happen if the Airseeker updates an underlying beacon separately, as part of the single beacon update flow

Exactly, this was the scenario that I was thinking about 🤭 I think it's best to remove it and let the beaconSet conditions check decide if the beaconSet update is necessary or not. I think I'm trying to optimize too soon here with that check 🤔

aquarat:
🎉 @ ✅

acenolaza:
> tada @ white_check_mark

The removal of the check in beaconSet update made even more sense when I went to fix the e2e 😁

aquarat:
Hey, can we merge this now 😅? (after the latest conflict is addressed of course)

acenolaza:
> Hey, can we merge this now sweat_smile? (after the latest conflict is addressed of course)

Sure thing. Let me rebase and fix the conficts and I'll merge it. Thanks for testing this 👏🏻

PR #253, opened 2022-09-28
Drop fetching blockNumber

bdrhn9:
blockNumber is only used while calling method `.getTransactionCount`. However, this method can be called without passing blockNumber explicitly. So, in order to reduce RPC calls, I removed redundant calls.


aquarat (approved):
LGTM 🔥 
This is a really nice find. I assume the original implementer just made a mistake/didn't realise `blockNumber`/`blockTag` defaulted to `latest`. Well done :)

PR #257, opened 2022-09-30
Allow Airseeker to make direct API call

### LINKED ISSUE #250: Allow Airseeker to directly retrieve API data
Currently Airseeker calls a remote HTTP signed data gateway in order to retrieve signed API responses. 

We'd like Airseeker to have the ability to call an API directly and sign the result with the existing Airseeker mnemonic.

This will practically create a hybrid between Airseeker and Airkeeper in terms of functionality.

This should be implemented as a `method` in the trigger configuration, for instance:
```json
  "triggers": {
    "dataFeedUpdates": {
      "1": {
        "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC": {
          "beacons": [
            {
              "beaconId": "0xbc80cbd7a8b8180e11d189a8334814a44a69c4d083b31305ecf67a3a3ea0fd9a",
              "deviationThreshold": 0.1,
              "heartbeatInterval": 86400
              "method": "direct"
            },
```
`method` should be capable of supporting other arbitrary retrievers through a common interface, eg.
```typescript
// something like...
interface DirectRequest {
  template: Id<Template>;
}

interface GatewayRequest extends DirectRequest {
  gateways: Gateway[];
}

interface Request {
  (request: DirectRequest | GatewayRequest): Promise<SignedData>
}

export const makeSignedDataGatewayRequests = async (
  gateways: Gateway[],
  template: Id<Template>
): Promise<SignedData> => {};
```

Methods could be:
- `direct` - call the API directly
- `v0.6.5` - call the remote gateway using an adapter for v0.6.5 Airnode
- `v0.9.0` - etc.

The above will require Airseeker to have access to the OIS. This could be included at the top level of `airseeker.json` as an array of OISes:
```json
{
  "triggers": {},
  "templates": {},
  "oises": [
    {
      "oisFormat": "1.1.1",
      "title": "<FILL_*>",
      "version": "<FILL_*>"
    }
  ]
}
```
Airseeker should search the available OISes for the first OIS that contains a matching endpoint and use the associated details to execute the direct call. The API response data should be signed with the Airseeker's mnemonic.

This will also require Airseeker to include an `apiCredentials` field in the config file.
Mnemonic derivation should be done up-front (like Airnode's deployer) to reduce duplication of processor usage.

### LINKED ISSUE #270: Change execution order of beacon update to reduce RPC calls
It'd better to mention about flow of updating beacon using Airseeker:

1. Get transactionCount (RPC call)
2. Read on-chain data (RPC call)
3. Check update conditions
4. If 3., Get gasPrice (RPC call)
5. Submit tx (RPC call)

For situations where the update condition isn’t satisfied (most frequent), we call RPC 2 times, if we can change order to 2-3-1-4-5, we’ll call RPC one time in that situation. 

This improvement doesn't look essential when only one sponsorWallet is used for beacons, because transaction count fetcher is invoked one times. But the situation where each beacon has distinct sponsorWallet, it effectively reduces the total number of RPC calls.

bdrhn9:
Solves #250. 

Additional to original issue, I added following features to save time instead of deriving sponsorWallets in each update cycle:
- `initializeAirseekerWallet` function that derives airseeker wallet (to sign encoded data on every API call) from mnemonic and saves it to `state`.
- Similarly, `initializeSponsorWallets` function that derives sponsor wallets (to sign data on every update) from mnemonic and saves them into `state`.


bdrhn9:
Made it draft again. After [this thread](https://api3workspace.slack.com/archives/C037NLX1R8T/p1664886937715659), it was decided to store private keys of wallets.

src/fetch-beacon-data.ts
```diff
@@ -65,25 +65,51 @@ export const fetchBeaconData = async (beaconId: string) => {
...
-    delay: { type: 'random', minDelayMs: RANDOM_BACKOFF_MIN_MS, maxDelayMs: RANDOM_BACKOFF_MAX_MS },
-    totalTimeoutMs: fetchInterval * 1_000,
-    onAttemptError: (goError) =>
-      logger.warn(`Failed attempt to call signed data gateway. Error: ${goError.error}`, logOptionsBeaconId),
-  });
-  if (!goRes.success) {
-    logger.warn(`Unable to call signed data gateway. Error: "${goRes.error}"`, logOptionsBeaconId);
-    return;
-  }
+  switch (method) {
```
acenolaza: Nit: I see quite a few repeated lines of code and I came up with this approach but it ended up being almost the same amount of lines 🤭 so I'm sharing it but please disregard if it doesn't add value:

```
export const fetchBeaconData = async (beaconId: string) => {
  const logOptionsBeaconId = { meta: { 'Beacon-ID': beaconId } };
  logger.debug('Fetching beacon data', logOptionsBeaconId);

  const { fetchInterval, airnode, templateId, method } = config.beacons[beaconId];
  const template = config.templates[templateId];

  let fetchFn: () => Promise<SignedData>;
  let onAttemptError: (goRes: GoResultError<Error>) => void;
  switch (method) {
    case 'direct': {
      fetchFn = () => makeDirectRequest({ ...template, id: templateId });
      onAttemptError = (goError: GoResultError<Error>) =>
        logger.warn(`Failed attempt to make direct call. Error: ${goError.error}`, logOptionsBeaconId);
      break;
    }
    case 'v0.6.5':
    case 'v0.9.0': {
      const gateway = config.gateways[airnode];
      fetchFn = () => makeSignedDataGatewayRequests(gateway, { ...template, id: templateId });
      onAttemptError = (goError: GoResultError<Error>) =>
        logger.warn(`Failed attempt to call signed data gateway. Error: ${goError.error}`, logOptionsBeaconId);
      break;
    }
    default:
      logger.warn(`Invalid API value fetch method ${method}`, logOptionsBeaconId);
      return;
  }

  const goRes = await go(fetchFn, {
    attemptTimeoutMs: GATEWAY_TIMEOUT_MS,
    retries: INFINITE_RETRIES,
    delay: { type: 'random', minDelayMs: RANDOM_BACKOFF_MIN_MS, maxDelayMs: RANDOM_BACKOFF_MAX_MS },
    totalTimeoutMs: fetchInterval * 1_000,
    onAttemptError,
  });
  if (!goRes.success) {
    logger.warn(`Unable to fetch beacon data. Error: "${goRes.error}"`, logOptionsBeaconId);
    return;
  }

  const { data } = goRes;
  if (data) {
    updateState((state) => ({ ...state, beaconValues: { ...state.beaconValues, [beaconId]: data } }));
  }
};
```

src/validation.ts
```diff
@@ -137,20 +148,76 @@ export const triggersSchema = z.object({
...
+    }
+
+    const endpoint = config.endpoints[template.endpointId];
+    if (isNil(endpoint)) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: `Endpoint "${template.endpointId}" is not defined in the config.endpoints object`,
+        path: ['templates', templateId, 'endpointId'],
+      });
+    }
```
acenolaza: Nit: isn't this easier to read 🤔 
```suggestion
    if (
      Object.values(config.beacons).some(({ templateId: tId, method }) => method === 'direct' && tId === templateId)
    ) {
      const endpoint = config.endpoints[template.endpointId];
      if (isNil(endpoint)) {
        ctx.addIssue({
          code: z.ZodIssueCode.custom,
          message: `Endpoint "${template.endpointId}" is not defined in the config.endpoints object`,
          path: ['templates', templateId, 'endpointId'],
        });
      }
    }
```

src/validation.ts
```diff
@@ -102,6 +106,13 @@ export const templatesSchema = z.record(config.evmIdSchema, templateSchema).supe
   });
 });
 
+export const endpointSchema = z.object({
+  oisTitle: z.string(),
+  endpointName: z.string(),
+});
```
acenolaza, line 112: Here you may also want to derive the `endpointId` and validate it like it's being done above for `templateIds`. See [here](https://github.com/api3dao/operations/blob/main/src/validation/validation-refinements.ts#L10-L12) for reference.

acenolaza: and add a test to validation.test.ts 😁

src/fetch-beacon-data.test.ts
```diff
@@ -11,27 +11,34 @@ const config: Config = {
...
       templateId: '0x9ec34b00a5019442dcd05a4860ff2bf015164b368cb83fcb756088fc6fbd6480',
       fetchInterval: 40,
+      method: 'v0.6.5',
     },
     '0x8fa9d00cb8f2d95b1299623d97a97696ed03d0e3350e4ea638f469beabcdabcd': {
       airnode: '0x5656D3A378B1AAdFDDcF4196ea364A9d78617290',
       templateId: '0x9ec34b00a5019442dcd05a4860ff2bf015164b368cb83fcb756088fcabcdabcd',
-      fetchInterval: 40,
+      fetchInterval: 0.5,
+      method: 'direct',
```
acenolaza: I'm assuming that the requirement is that this is defined on per beacon basis right? 

Also the name `method` might be a bit too generic and I get the feeling that if I look at this property in a couple of months, I'll have no idea what is it for? 😁 Maybe `fetchMethod` or `fetchApiValueFrom`, etc. The values could be `api` instead of direct and something like `gateway0.9.0` (I'm really bad at naming things 🤣 )

src/make-request.ts
```diff
@@ -85,3 +100,44 @@ export const makeSignedDataGatewayRequests = async (
...
+  const parameters: node.ApiCallParameters = abi.decode(template.parameters);
+  const endpoint: Endpoint = config.endpoints[template.endpointId];
+
+  const aggregatedApiCall: node.BaseAggregatedApiCall = {
+    parameters,
+    ...endpoint,
+  };
+  const [_, apiCallResponse] = await node.api.callApi({
+    type: 'http-gateway',
+    config,
```
acenolaza: Here I would just pass what's required by `node.api.callApi()`: 
```suggestion
    config: { ois, apiCredentials },
```
You'll just need to destructure the `ois`, `apiCredentials` objects from config like this:

```
  const {
    config: { endpoints, ois, apiCredentials },
  } = getState();
```

src/update-data-feeds.ts
```diff
@@ -261,7 +256,11 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
 
     // Get the latest gas price
     const [logs, gasTarget] = await getGasPrice(provider.rpcProvider, config.chains[chainId].options);
-    logs.forEach((log) => (log.level === 'ERROR' ? logger.error(log.message) : logger.info(log.message)));
+    logs.forEach((log) =>
+      log.level === 'ERROR'
+        ? logger.error(`${log.message}-${provider.providerName}`)
+        : logger.info(`${log.message}-${provider.providerName}`)
```
acenolaza: Maybe instead of concatenating the `providerName` you could pass the `logOptionsBeaconId` object that will contain provider+chain+beacon metadata

acenolaza (approved):
Overall I really like these changes 🥇  and I just left a couple of comments/suggestions that you may or may not use.

src/fetch-beacon-data.ts
bdrhn9: I think your touch looks better :)

src/fetch-beacon-data.test.ts
bdrhn9: Actually, in related issue [#250](https://github.com/api3dao/airseeker/issues/250), `method` is specified under `config.triggers.datafeedUpdates.<chainId>.<sponsorWallet>.beacons[idx].method`. However, defining method keyword under triggers causes changing fetching architecture dramatically, because all required information for fetchBeaconData is under config.beacons as can be seen [here](https://github.com/api3dao/airseeker/blob/e4f19b7505d91e98c66fbdfd816ab56eeac12afa/src/fetch-beacon-data.ts#L68). Also triggers sounds like updating beacon related.

Welll 😸, I didn't make a decision about namings. Should I do these changes or maybe we can discuss on Slack? I'm totally up-to you.

bdrhn9:
> Overall I really like these changes 1st_place_medal and I just left a couple of comments/suggestions that you may or may not use.

Happy to see that comment. Thank you 🔥

src/fetch-beacon-data.test.ts
amarthadan: As for the naming issue, I agree with what @acenolaza suggested. I'm ok with the key being called `fetchMethod` and possible values being named `api`, `gateway-old` and `gateway` (as this will be the default for the foreseeable future). WDYT?

As for where to put the `method` field, I presume putting it under `triggers` was to allow using the same beacons in different configurations. Meaning you would be able to have the same beacon being fetched both via a gateway and directly. If you add the field under the topmost `beacons`, you won't be able to do that, and the given beacon will be bound to a specific fetch method. @aquarat Correct me if I'm wrong. I'm not sure how much flexibility is needed here.

src/fetch-beacon-data.ts
```diff
@@ -65,20 +66,40 @@ export const fetchBeaconData = async (beaconId: string) => {
...
+  let onAttemptError: (goRes: GoResultError<Error>) => void;
+  switch (method) {
+    case 'direct': {
+      fetchFn = () => makeDirectRequest({ ...template, id: templateId });
+      onAttemptError = (goError: GoResultError<Error>) =>
+        logger.warn(`Failed attempt to make direct call. Error: ${goError.error}`, logOptionsBeaconId);
+      break;
+    }
+    case 'v0.6.5':
+    case 'v0.9.0': {
```
amarthadan: Why do we even have this distinction? It's running the same code as far as I can tell.

src/validation.ts
```diff
@@ -102,6 +106,30 @@ export const templatesSchema = z.record(config.evmIdSchema, templateSchema).supe
...
+
+    if (derivedEndpointId !== endpointId) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: `Endpoint ID "${endpointId}" is invalid`,
+        path: [endpointId],
+      });
+    }
+  });
+});
```
amarthadan, line 131: I know we don't do this check in Airnode because it's more of a convention than a rule. Is it a rule here?

amarthadan (commented):
I think the position of the `method` key within the config should be agreed on prior to merging @aquarat 
Other than that, just some small things.

src/fetch-beacon-data.test.ts
bdrhn9: Yes you are right, current approach restricts fetching via both methods. Okay, I can quickly give a try for changing `fetchMethod` position into triggers section 👍🏼 Also I like the `gateway-old` and `gateway` namings and gonna apply them.

bdrhn9: The problem with new positioning `fetchMethod` is that while directly calling API Airseeker behaves like an Airnode (fetches data via OIS and templateId then signs the data) so [here](https://github.com/api3dao/airseeker/blob/1f2078ccd6e727ff5a6fbe72e62e7f9cdba80184/config/airseeker.example.json#L21) `airnode` should be the address of this Airseeker which will be part of Airnode v.1.*. In this situation, we need to add same key (beacon id) twice into `config.beacons` object which is not possible.

src/fetch-beacon-data.ts
bdrhn9: You're right. This distinction can be used to select parsing [here](https://github.com/api3dao/airseeker/blob/1f2078ccd6e727ff5a6fbe72e62e7f9cdba80184/src/make-request.ts#L69). Other than that it's not necessary. I think it's better to left this distinction someone who's going to implement it in future 😄

src/validation.ts
bdrhn9: Idk actually. Maybe @acenolaza can help me to answer this question.

src/fetch-beacon-data.test.ts
aquarat: >  @aquarat Correct me if I'm wrong. I'm not sure how much flexibility is needed here.

My understanding is that Airseeker will be operated (potentially) in two different ways, generally associated with two different operating _parties_:
- A first party operator (the API provider, in this mode Airseeker behaves like Airkeeper)
- A third party operator (eg. "Us")

In the third party case, the way Airseeker is currently used, the data is signed by the remote Airnode and handed off to the `DapiServer`, which then verifies the signature. In this situation the mnemonic/key being used to execute the transaction is immaterial. In these situations you'd always want the data signed by the remote gateway, never directly.

In the first party case, you'd always want to sign the data using the local key, not a remote Airnode, because you, as the API provider, are supplying the data so it should carry your signature.

So as I understand it these two situations wouldn't mix and therefore it's acceptable for the method to be on a per beacon basis 🤷

src/validation.ts
acenolaza: Well, @amarthadan rises a good point here. I remember seeing this validation in operations repo and I think @aquarat added it because Airnode config.json was being auto-generated and it used [this function](https://github.com/api3dao/airnode/blob/master/packages/airnode-admin/src/implementation.ts#L258) to derive the `endpointIds`. So it feels like it's more of a rule now but I can't say this for sure.

src/fetch-beacon-data.test.ts
amarthadan: In that case, let's keep the `fetchMethod` field where it is :slightly_smiling_face:

src/validation.ts
amarthadan: Makes sense, I just wanted to make sure. I don't think it's a bad thing to force it.

src/fetch-beacon-data.ts
amarthadan: I see. Can you create an issue for that so we don't forget about it?

PR #264, opened 2022-10-07
Allow creation of airseeker configuration 

Ashar2shahid:
part of refactoring operations is to migrate the creation scripts away from operations and into the respective repositories. With this script it should be easy to create airseeker config scripts based on operations data.


scripts/create-airseeker-config.ts
```diff
@@ -0,0 +1,295 @@
...
+          })).reduce((r, c) => Object.assign(r, c), {}),
+          options: {
+            txType: 'eip1559' as const,
+            priorityFee: {
+              value: 3.12,
+              unit: 'gwei' as const,
+            },
+            baseFeeMultiplier: 2,
+            fulfillmentGasLimit: 500000,
+          },
```
bdrhn9: I think that this changes is required to satisfy latest validation schema 😸 
```suggestion
          options: {
            fulfillmentGasLimit: 500000,
            gasPriceOracle: [
              {
                gasPriceStrategy: 'latestBlockPercentileGasPrice',
                percentile: 60,
                minTransactionCount: 20,
                pastToCompareInBlocks: 20,
                maxDeviationMultiplier: 2,
              },
              {
                gasPriceStrategy: 'providerRecommendedGasPrice',
                recommendedGasPriceMultiplier: 1.2,
              },
              {
                gasPriceStrategy: 'constantGasPrice',
                gasPrice: {
                  value: 10,
                  unit: 'gwei',
                },
              },
            ],
          },
```

package.json
```diff
@@ -21,6 +21,7 @@
     "clean": "rimraf -rf ./build ./dist *.tgz",
     "compile": "tsc --build tsconfig.json",
     "dev:create-local-config": "ts-node scripts/create-local-config.js",
+    "dev:create-operations-config": "ts-node scripts/create-airseeker-config.js",
```
bdrhn9:
```suggestion
    "dev:create-operations-config": "ts-node scripts/create-airseeker-config.ts",
```

scripts/create-airseeker-config.ts
```diff
@@ -0,0 +1,295 @@
...
+      ...operationsRepository.api3,
+      airseeker: {
+        ...operationsRepository.api3?.airseeker,
+        [response.name]: {
+          airseeker,
+          secrets,
+        },
+      },
+    },
+  };
```
bdrhn9: `updatedOpsData` is not used later so it can be deleted. Btw, it would be better to have deployment folder imo.

scripts/create-airseeker-config.ts
```diff
@@ -0,0 +1,295 @@
...
+import { getFormattedUtcTimestamp } from './utils';
+
+const questions = (beaconChoices: Choice[], beaconSetChoices: Choice[]): PromptObject[] => {
+  return [
+    {
+      type: 'text',
+      name: 'name',
+      message: 'What is the Airseeker configuration name?',
+      initial: getFormattedUtcTimestamp(),
+    },
```
bdrhn9: If `updatedOpsData` is going to be deleted, this question will be also redundant.

scripts/create-airseeker-config.ts
```diff
@@ -0,0 +1,295 @@
...
+      type: 'number',
+      name: 'numberOfProviders',
+      message: 'How many RPC provider (per chain) you want to use in the configuration?',
+      validate: (value) => (value < 0 ? 'Non-negative values are not allowed!' : true),
+    },
+  ];
+};
+
+const main = async (operationRepositoryTarget?: string) => {
+  const operationsRepository = readOperationsRepository(operationRepositoryTarget);
```
bdrhn9: Maybe another question can be added that asks `operationRepositoryTarget` for creating config from whether remote or local files. Totally optional 😸

bdrhn9 (changes requested):
After zod validator satisfied, LGTM 🔥

scripts/create-airseeker-config.ts
Ashar2shahid: yeah I added this for airkeeper but I was wondering if its needed for airseeker. I added it in with the other PR fixes

aquarat:
If @bdrhn9 is happy I'm happy (thanks for reviewing). Although as you noted some of this work will be made redundant by the ops repo migration to a database.

scripts/create-airseeker-config.ts
```diff
@@ -0,0 +1,316 @@
...
+          Math.min(
+            ...Object.values(beacon.chains)
+              .filter((chain) => 'airseekerConfig' in chain)
+              .map((chain) => chain.airseekerConfig!.updateInterval)
+          ) / 2
+        ),
+      },
+    }),
+    {} as Beacons
+  );
```
bdrhn9:
```suggestion
  const airseekerBeacons = beacons.reduce(
    (beaconObj, beacon) => ({
      ...beaconObj,
      [beacon.beaconId]: {
        airnode: beacon.airnodeAddress,
        templateId: beacon.templateId,
        fetchInterval: Math.ceil(
          Math.min(
            ...Object.values(beacon.chains)
              .filter((chain) => 'airseekerConfig' in chain)
              .map((chain) => chain.airseekerConfig!.updateInterval)
          ) / 2
        ),
        fetchMethod: 'gateway',
      },
    }),
    {} as Beacons
  );
```

scripts/create-airseeker-config.ts
```diff
@@ -0,0 +1,316 @@
...
+  const airseeker = {
+    airseekerWalletMnemonic: '${AIRSEEKER_WALLET_MNEMONIC}',
+    log: airseekerLogs,
+    beacons: airseekerBeacons,
+    beaconSets: airseekerBeaconSet,
+    chains: airseekerChains,
+    gateways: airseekerGateways,
+    templates: aiseekerTemplates,
+    triggers: airseekerTriggers,
+  };
```
bdrhn9:
```suggestion
  const airseeker = {
    airseekerWalletMnemonic: '${AIRSEEKER_WALLET_MNEMONIC}',
    log: airseekerLogs,
    beacons: airseekerBeacons,
    beaconSets: airseekerBeaconSet,
    chains: airseekerChains,
    gateways: airseekerGateways,
    templates: aiseekerTemplates,
    triggers: airseekerTriggers,
    endpoints: {},
    ois: [],
    apiCredentials: [],
  };
```

bdrhn9 (approved):
After merging #257, latest suggestions needs to be committed for backward compatibility. I tested this changes, feel free to commit them 👍🏼 
As a note, with these suggestions, this scripts only works for fetching via gateways. For newly added direct call method, functions that prepare `endpoints`, `ois`, and `apiCredentials` needs to be added. I think it's going to be easy, because we already implemented them for Airkeeper. Issue can be created for this situation.

Ashar2shahid:
I didn't realize I hadn't merged this yet 🤦 . Applied the PR suggestions that bedrihan suggested and he's right we can create new issue to add the new endpoints

PR #276, opened 2022-10-12
Change execution order of beacon update to reduce RPC calls

### LINKED ISSUE #270: Change execution order of beacon update to reduce RPC calls

bdrhn9:
Solves #270.


src/update-data-feeds.ts
```diff
@@ -245,6 +222,20 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
...
     }
 
+    // Get transaction count only first time when update condition satisfied
+    if (isNil(nonce)) {
+      const transactionCount = await getTransactionCount(
+        provider,
+        sponsorWallet.address,
+        prepareGoOptions(startTime, totalTimeout)
+      );
+      if (transactionCount === null) {
```
acenolaza: Nit:
```suggestion
    if (!nonce) {
      const transactionCount = await getTransactionCount(
        provider,
        sponsorWallet.address,
        prepareGoOptions(startTime, totalTimeout)
      );
      if (isNil(transactionCount)) {
```

acenolaza (approved):
LGTM 🥇  We might just want to add back a test on `describe("updateBeaconSets")` when transaction count cannot be found tho (not sure why we don't have a `describe("updateBeacons")` 🤔 )

PR #291, opened 2022-10-21
Enable building the project with production dependencies

vponline:
Required for deployment with operations-database airseeker deployer lambda where it uses `npm install --production --ignore-scripts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,8 @@
     "lodash": "^4.17.21",
     "promise.any": "^2.0.4",
     "source-map-support": "^0.5.21",
+    "serverless": "^3.23.0",
+    "typescript": "^4.8.4",
     "zod": "^3.17.3"
   },
   "devDependencies": {
@@ -75,9 +77,7 @@
     "pm2": "^5.2.2",
     "prettier": "^2.7.1",
     "rimraf": "^3.0.2",
-    "serverless": "^3.23.0",
     "ts-jest": "^28.0.8",
-    "ts-node": "^10.9.1",
-    "typescript": "^4.8.4"
+    "ts-node": "^10.9.1"
   }
 }


package.json
```diff
@@ -51,6 +51,8 @@
     "lodash": "^4.17.21",
     "promise.any": "^2.0.4",
     "source-map-support": "^0.5.21",
+    "serverless": "^3.23.0",
+    "typescript": "^4.8.4",
```
acenolaza, line 55: May I ask why typescript needs to be packed for prod? As I understand serverless only uploads a zip file with all the JS files inside, right? Also, have you tried actually deploying this to AWS? I remember having zip file limitation issues with Airkeeper when adding too many deps.

acenolaza: I just saw this comment https://github.com/api3dao/operations-database/pull/46/files#r1001427998

vponline: > May I ask why typescript needs to be packed for prod? As I understand serverless only uploads a zip file with all the JS files inside, right?

Yeah definitely a good question. Just for a bit more context, we need typescript in the production dependencies when installing with the `--production` flag so that `sls deploy` can run `tsc` to compile. I wasn't able to install the dev dependencies on AWS lambda and these two seem to be the only ones missing to use `--production`. Any other suggestions are definitely welcome as this may not be the best approach.

> Also, have you tried actually deploying this to AWS? I remember having zip file limitation issues with Airkeeper when adding too many deps.

Yep I was able to deploy an Airseeker from a lambda with this change, although I haven't tested if that Airseeker would actually work as expected.

acenolaza: > Yeah definitely a good question. Just for a bit more context, we need typescript in the production dependencies when installing with the --production flag so that sls deploy can run tsc to compile. I wasn't able to install the dev dependencies on AWS lambda and these two seem to be the only ones missing to use --production. Any other suggestions are definitely welcome as this may not be the best approach.

Right, I think I got it after looking at the operations-database PR but thanks anyways for clarifying this further. I guess the main issue is that we are not building and deploying Airseeker in a somewhat controlled environment since the code is cloned and compiled in lambda function. My instinct tells me that docker should be used here but tbh I'd appreciate @amarthadan feedback on this topic.

vponline: Yeah this is definitely quite a hack (the idea was to save dev time if it happened to work).

A more _official_ approach would be to use the AWS SDK's [CreateFunction](https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html) method, which can take a ZIP or a docker image (from Amazon ECR). Can we run docker inside of a Lambda :thinking:? Perhaps the Airseeker CI should build the docker container and upload to ECR?

amarthadan: Just to make sure I understand what you're trying to do. You're using lambda to clone an Airseeker repo and deploy the Airseeker from that lambda, right?

This is definitely not the correct way to do this. The app's dependencies shouldn't change for the sake of the environment. Airseeker doesn't need neither serverless, nor typescript to run (prod dependencies). It needs them to be built and deployed (dev dependencies). That's why there's this division. The dev dependencies should be installed in the environment doing the build and deployment.

Not sure how your operations build environment is set up but you would have a much simpler time running these builds and deployments from the Docker container as @acenolaza mentioned. Also not sure what else will be deployed from there but I presume you'll run into a similar issue if trying to deploy Airnode itself.

I would suggest either using a custom lambda environment with whatever you need (essentially a Docker image as that's what lambdas are) or one-off ECS FARGATE containers. But I would definitely not recommend what this PR is doing.

vponline: > Just to make sure I understand what you're trying to do. You're using lambda to clone an Airseeker repo and deploy the Airseeker from that lambda, right?

Yep, doing the deployment with serverless from inside the lambda :sweat_smile:

Should we consider these alternatives (Perhaps even [AWS CodeBuild](https://aws.amazon.com/codebuild/) could be suitable where the build can be triggered by the lambda)?

acenolaza: @amarthadan as always, very valuable information. Thanks for the input!
@vponline I'm very sorry for over-complicating things but I looked weird to have to add a bunch of dev dependencies to the built package. My main concern now is zip file and having to probably apply same hack for telemetry and other apps deployed using this approach. At the same time I know time and resources is limited and stuff to do is piling up. So if @aquarat is OK with this approach then we can continue with this and create a ticket for improving it in the future. I just wouldn't like to end up reinventing k8s or something 😅

vponline: No worries, your input is always welcome (and this approach was admittedly a quick-fix to just get it working on my part) :smile:.

> My main concern now is zip file and having to probably apply same hack for telemetry and other apps deployed using this approach. 

Ah yes, I will have to see about telemetry when I get there :sweat_smile: 

This change won't be merged and we will instead publish Airseeker as a package and see if that works better. I will close this PR for now.

PR #298, opened 2022-10-28
Require 1, but default to 3, airseeker config RPC providers

### LINKED ISSUE #263: No default provider amount when no input is provided
When you run `create-airseeker-config` it asks

> How many RPC provider (per chain) you want to use in the configuration?

If you hit enter, the providers are empty. Since there is no point in having a chain without any providers for an Airseeker it should ask this instead

> How many RPC provider (per chain) you want to use in the configuration? (default: 3)

dcroote:
Closes #263. The first commit fixes a zod-related type error.

### DIFF
--- a/scripts/create-airseeker-config.ts
+++ b/scripts/create-airseeker-config.ts
@@ -45,8 +45,9 @@ const beaconSetQuestions = (beaconChoices: Choice[], beaconSetChoices: Choice[])
     {
       type: 'number',
       name: 'numberOfProviders',
-      message: 'How many RPC provider (per chain) you want to use in the configuration?',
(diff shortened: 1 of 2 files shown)


acenolaza:
🤔 so the plan is to listen for changes on rows to some tables in operations-database postgresql. If that happens then a new airseeker.json file is generated and pushed to a repo. Then when that PR is merged, Airseeker will automagically be deployed using the updated config file. With this in mind, should we even keep this script here in airseeker repo @aquarat ?

aquarat (approved):
There's nothing wrong with this but this script is about to be made redundant due to the migration of operations to the operations-database (as @acenolaza alluded to). The source issue was unfortunately out of date, I'll have to do some filtering asap.

PR #319, opened 2022-11-30
Add ncc, remove operations dependency, various minor fixes

aquarat:
This PR was largely created out of needs discovered while trying to make this app work with [the airseeker-deployer](https://github.com/api3dao/operations-database/pull/76).

This PR:
- Adds Vercel's ncc as a build step. This is similar to webpack. It reduces the serverless zipped payload size from ~75 MB down to 5 MB. The build output generated by the deployer worker, unzipped, would sometimes exceed the Lambda size limits, adding ncc fixes that.
- Removes the `create-airseeker-config.ts` script. This script relies on operations, which weirdly added circa 100 MB to the serverless ZIP payload. The `operations` repository is also deprecated.
- Associated changes to the example serverless file (which the deployer uses with interpolation).


aquarat:
Note to self (@aquarat) add documentation to README

vponline (approved):
LGTM, ncc is pretty amazing :smile:

aquarat:
Actually it doesn't look like documentation for ncc is necessary; it's just part of the existing build step. So I'll leave that.

PR #333, opened 2023-01-12
327 fix dapi server interface changes

### LINKED ISSUE #327: Update signed data updates according to the changes in the DapiServer interface
DapiServer signed data update interface changed a bit. Airseeker should be updated accordingly. There are two ways of doing this
- Do a breaking change
- Add a new trigger type that uses the new interface for backward/forward compatibility

acenolaza:
Closes #327 

 ~This is a draft PR because this requires that a new version of the airnode-protocol-v1 package is published and then we can update the dependencies here.~


bbenligiray (approved):

vponline (approved):
LGTM :medal_sports:

PR #338, opened 2023-01-18
Read multiple data feeds with a single static call #256

### LINKED ISSUE #256: Read multiple data feeds with a single static call
Similar to https://github.com/api3dao/airseeker/issues/216

An Airseeker that runs thousands of Beacons needs to make a lot of RPC calls to read these individually. It would be better to read these in batches through Multicall.

acenolaza:
Closes #256 

Draft PR because I'd like to run Airseeker locally on a couple of chains and confirm that everything is still working as expected

### DIFF
--- a/src/update-data-feeds.ts
+++ b/src/update-data-feeds.ts
@@ -11,8 +11,8 @@ import {
   checkUpdateCondition,
 } from './check-condition';
 import {
-  BEACON_READ_BATCH_SIZE,
-  BEACON_UPDATE_BATCH_SIZE,
+  DATAFEED_READ_BATCH_SIZE,
+  DATAFEED_UPDATE_BATCH_SIZE,
   INT224_MAX,
   INT224_MIN,
   NO_DATA_FEEDS_EXIT_CODE,
@@ -159,7 +159,8 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
 
   type BeaconUpdateData = {
     logOptionsBeaconId: LogOptionsOverride;
-    beaconUpdateData: BeaconUpdate & Beacon;
+    beaconUpdate: BeaconUpdate;
+    beacon: Beacon;
     newBeaconResponse: SignedData;
     newBeaconValue: ethers.BigNumber;
     dataFeedsCalldata: string;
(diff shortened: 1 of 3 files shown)


src/update-data-feeds.ts
```diff
@@ -181,68 +192,117 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
...
+        logOptionsBeaconId,
+        beaconUpdateData,
+        newBeaconResponse,
+        newBeaconValue,
+        calldata: contract.interface.encodeFunctionData('readDataFeedWithId', [beaconUpdateData.beaconId]),
+      },
+    ];
+  }
+
+  // TODO: should this also be batched? it's just a static call so it might be ok...
```
bbenligiray: We were talking about this @bdrhn9 the other day (specifically about multicall-ing sponsor wallet balance checks, which is a comparable operation). The problem with static calls with very large gas costs is that they will take longer and may time out, or be blocked by the providers. A batch size of 100 would be safer.

vponline: > The problem with static calls with very large gas costs is that they will take longer and may time out, or be blocked by the providers. A batch size of 100 would be safer.

If datafeed value reading starts failing because of this (maybe the RPC provider changed something), will we know about it (besides seeing it in the logs and updates not happening)? We'd need a quick fix to go back to the old single read method, right?

vponline (approved):

src/update-data-feeds.ts
acenolaza: Or we might want to take Aaron's approach of using a env var instead of a const so we can change the batch size without having to re-deploy Airseeker 🤔

acenolaza:
~This might require more changes after [this discussion](https://api3workspace.slack.com/archives/C02AYRX8D89/p1673448962958209) comes to a conclusion.~

UPDATE: discussion suggested to replace `DapiServer.readDataFeedWithId()` call with `DapiServer.datafeeds()` to avoid revert if data feed hasn't been initialized yet.

src/update-data-feeds.ts
bbenligiray: I also mentioned @bdrhn9 that being able to make high gas limit static calls should be a part of our eligibility criteria for blockchain providers

acenolaza:
Merged this branch with the branch that also contains the changes for tryMulticall updates because both are closely related and makes my life easier with merge conflicts and switching between the branches.

PR #341, opened 2023-01-20
Instead of skipping the beacon if datafeed has not been initialized w…

acenolaza:
…e skip the checks instead and perform the signed update

### DIFF
--- a/src/update-data-feeds.ts
+++ b/src/update-data-feeds.ts
@@ -185,41 +185,45 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
       voidSigner,
       contract,
       beaconUpdateData.beaconId,
-      prepareGoOptions(startTime, totalTimeout),
+      { ...prepareGoOptions(startTime, totalTimeout), retries: 2 },
       logOptionsBeaconId
     );
     if (!onChainData) {
-      continue;
-    }
-
-    // Check that signed data is newer than on chain value
-    const isSignedDataFresh = checkBeaconSignedDataFreshness(onChainData.timestamp, newBeaconResponse.timestamp);
-    if (!isSignedDataFresh) {
-      logger.warn(`Signed data older than on chain record. Skipping.`, logOptionsBeaconId);
-      continue;
-    }
-
-    // Check that on chain data is newer than heartbeat interval
-    const isOnchainDataFresh = checkOnchainDataFreshness(onChainData.timestamp, beaconUpdateData.heartbeatInterval);
-    if (!isOnchainDataFresh) {
-      logger.info(
-        `On chain data timestamp older than heartbeat. Updating without condition check.`,
-        logOptionsBeaconId
-      );
+      logger.warn('🚀 DATA FEED NOT INITIALIZED!!!', logOptionsBeaconId);
     } else {
-      // Check beacon condition
-      const shouldUpdate = checkUpdateCondition(onChainData.value, beaconUpdateData.deviationThreshold, newBeaconValue);
-      if (shouldUpdate === null) {
-        logger.warn(`Unable to fetch current beacon value`, logOptionsBeaconId);
-        // This can happen only if we reach the total timeout so it makes no sense to continue with the rest of the beacons
-        return;
-      }
-      if (!shouldUpdate) {
-        logger.info(`Deviation threshold not reached. Skipping.`, logOptionsBeaconId);
+      // Check that signed data is newer than on chain value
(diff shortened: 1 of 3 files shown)


bdrhn9:
The only thing on this implementation doubts me is when reading on-chain data fails because of some rpc problem. If data feed is already initialized, this'll update on-chain value without checking any condition maybe it doesn't exceed the threshold so it burnt gas unnecessarily. Other than that, it looks safe.

How about changing only `read-data-feed-with-id.ts` to mimic previous dapiServer answer (returns 0 for both value and timestamp for uninitialized data feed). I tried both your implementation and mine, both of them works okay. Here is differences for suggested approach:

```diff --git a/src/read-data-feed-with-id.ts b/src/read-data-feed-with-id.ts
index f4cdd39..c04a817 100644
--- a/src/read-data-feed-with-id.ts
+++ b/src/read-data-feed-with-id.ts
@@ -8,6 +8,10 @@ export interface DataFeed {
   timestamp: number;
 }
 
+const nullValue = ethers.BigNumber.from('0x00');
+const nullTimestamp = 0;
+const uninitializedDataFeed = Object.assign([nullValue, nullTimestamp], { value: nullValue, timestamp: nullTimestamp });
+
 export const readDataFeedWithId = async (
   voidSigner: ethers.VoidSigner,
   dapiServer: DapiServer,
@@ -22,10 +26,13 @@ export const readDataFeedWithId = async (
 
   const goDataFeed = await go(() => dapiServer.connect(voidSigner).readDataFeedWithId(dataFeedId), {
     ...goOptions,
-    onAttemptError: (goError) =>
-      logger.warn(`Failed attempt to read data feed. Error: ${goError.error}`, logOptionsDapiServerAddress),
   });
+
   if (!goDataFeed.success) {
+    if ((goDataFeed.error as any).reason && (goDataFeed.error as any).reason === 'Data feed not initialized') {
+      logger.warn('🚀 DATA FEED NOT INITIALIZED!!!', logOptionsDapiServerAddress);
+      return uninitializedDataFeed;
+    }
     logger.warn(`Unable to read data feed. Error: ${goDataFeed.error}`, logOptionsDapiServerAddress);
     return null;
   }
```

Here is properly highlighted code for `read-data-feed-with-id.ts`:
```typescript
import { DapiServer } from '@api3/airnode-protocol-v1';
import { GoAsyncOptions, go } from '@api3/promise-utils';
import { ethers } from 'ethers';
import { logger, LogOptionsOverride } from './logging';

export interface DataFeed {
  value: ethers.BigNumber;
  timestamp: number;
}

const nullValue = ethers.BigNumber.from('0x00');
const nullTimestamp = 0;
const uninitializedDataFeed = Object.assign([nullValue, nullTimestamp], { value: nullValue, timestamp: nullTimestamp });

export const readDataFeedWithId = async (
  voidSigner: ethers.VoidSigner,
  dapiServer: DapiServer,
  dataFeedId: string,
  goOptions: GoAsyncOptions,
  logOptions: LogOptionsOverride
): Promise<DataFeed | null> => {
  const logOptionsDapiServerAddress = {
    ...logOptions,
    meta: { ...logOptions.meta, 'Dapi-Server': dapiServer.address },
  };

  const goDataFeed = await go(() => dapiServer.connect(voidSigner).readDataFeedWithId(dataFeedId), {
    ...goOptions,
  });

  if (!goDataFeed.success) {
    if ((goDataFeed.error as any).reason && (goDataFeed.error as any).reason === 'Data feed not initialized') {
      logger.warn('🚀 DATA FEED NOT INITIALIZED!!!', logOptionsDapiServerAddress);
      return uninitializedDataFeed;
    }
    logger.warn(`Unable to read data feed. Error: ${goDataFeed.error}`, logOptionsDapiServerAddress);
    return null;
  }

  return goDataFeed.data;
};

```

acenolaza:
> it burnt gas unnecessarily

Yes, this might be an issue on mainnet. I don't know on other chains how much of an issue that would be. I guess it would depend on how long it would take us to find out about the RPC (or other) issue or how frequent an issue like that could happen.

> How about changing only `read-data-feed-with-id.ts` to mimic previous dapiServer answer

I personally avoid as much as possible writing logic that depends on comparing strings but I guess I don't have a strong case against doing that this time. So I guess it might be fine to check the revert string after making all lowercase before comparing. [Here](https://github.com/api3dao/airseeker/pull/338) when using `tryMulticall` to call `readDatafeedValueWithId` we will probably need to decode the revert string (I've seen a helper function to do this decoding somewhere, probably airnode repo or airnode-protocol-v1 test). Thanks for figuring this out! 👏🏻

PR #342, opened 2023-01-23
Read data feed values using the getter function instead of readDataFe…

acenolaza:
…edWithId function in DapiServer


src/update-data-feeds.ts
```diff
@@ -181,14 +180,19 @@ export const updateBeacons = async (providerSponsorBeacons: ProviderSponsorDataF
...
+      ...prepareGoOptions(startTime, totalTimeout),
+      onAttemptError: (goError) =>
+        logger.warn(`Failed attempt to read data feed. Error: ${goError.error}`, logOptionsBeaconId),
+    });
+    if (!goDataFeed.success) {
+      logger.warn(`Unable to read data feed. Error: ${goDataFeed.error}`, logOptionsBeaconId);
+      continue;
+    }
+    const onChainData = goDataFeed.data;
+    if (isEmpty(onChainData) || onChainData.timestamp === 0) {
```
bdrhn9: Why are we skipping if data feed isn't initialized?

acenolaza: Good catch! I was trying out something with the e2e tests and forgot to remove the right side of the `||`

bdrhn9 (approved):
LGTM 🔥

PR #376, opened 2023-05-04
PoC Airseeker fixes for beacon sets

aquarat:
https://api3workspace.slack.com/archives/C040G0W7RH9/p1683128039798789?thread_ts=1683100578.835749&cid=C040G0W7RH9

@acenolaza 

This branch splits out the main component of the single-beacon update function into a discrete function. It then calls that function for every beacon inside a beaconset to try and update those beacons first before attempting to update the parent beacon set. This prevents a scenario where Airseeker tries to update a beacon set using off-chain data but the transaction fails because the on-chain data is stale/doesn't match. 

This branch also uses the aggregation function to aggregate the timestamps (median), which is what the contract does (a relatively recent change).

LMK if you need the config file on Keybase.

For efficiency we probably need to make the updateBeacon function accept a "checkOnly" arg, run through the feeds and if more than one needs to be updated, use multicall otherwise single update.


acenolaza:
> This branch splits out the main component of the single-beacon update function into a discrete function. It then calls that function for every beacon inside a beaconset to try and update those beacons first before attempting to update the parent beacon set. This prevents a scenario where Airseeker tries to update a beacon set using off-chain data but the transaction fails because the on-chain data is stale/doesn't match.

I'm pretty sure that updating the `beaconSet` constituents was the original plan but because it was too expensive we ended up not doing it. @bbenligiray or @amarthadan might remember

BTW I would probably not use the new updateBeacon function to update the beaconSet beacons since we don't really want to go thru all the validation that the function does. It feels like it should be more of a force update (calling the contract directly for each beacon), right?

> This branch also uses the aggregation function to aggregate the timestamps (median), which is what the contract does (a relatively recent change).

I'm quite confused about this check 😁 This is the same check that is causing the revert mentioned above and we are trying to update the beaconSet beacons in order to get past that require statement in the contract. So wouldn't this check in Airseeker prevent it from updating the beaconSet beacons causing the same end result of the beaconSet not getting updated because of stale beacons?

> LMK if you need the config file on Keybase.

I would definitely like to run this locally to be able to debug the code. Also multicall PR should indeed help a lot with gas usage if Airseeker is configured to use just a few sponsor wallets. The multicall PR is not currently updating the beaconSet beacons and it's probably also lacking the median(onchainBeaconTimestamps).

aquarat:
> I'm pretty sure that updating the beaconSet constituents was the original plan but because it was too expensive we ended up not doing it. @bbenligiray or @amarthadan might remember

🙈 😅 
 It looks like the function that enabled updating a beacon set with signed data was removed from airnode-protocol-v1. In other words, in the current implementation that only way of updating a beacon set appears to be by first updating the constituents. AFAIK this is now the approach.

Assuming the above is correct, it then becomes a question of how and when those child data feeds should be updated; multicall, definitely, but do we update them with their existing-configured update triggers or do we perhaps do a more advanced check where we try to determine if updating a child would have a material effect on the parent?

Like maybe iterate through children, for every child check if the beaconset should be updated with all feeds excluding that child, if no, update it as part of the main update.

> I'm quite confused about this check 😁 This is the same check that is causing the revert mentioned above and we are trying to update the beaconSet beacons in order to get past that require statement in the contract. So wouldn't this check in Airseeker prevent it from updating the beaconSet beacons causing the same end result of the beaconSet not getting updated because of stale beacons?

The median check for timestamps was discussed a month or two back I think. In the original implementation ([see here](https://github.com/api3dao/airnode-protocol-v1/blob/dbf7564546e70bdebdb118c50230b19f98f2bc4e/contracts/dapis/DapiServer.sol#L1000)) it was an average of the submitted timestamps. Naturally the average would be unlikely to match an equality check, so there was [a greater-than check instead](https://github.com/api3dao/airnode-protocol-v1/blob/dbf7564546e70bdebdb118c50230b19f98f2bc4e/contracts/dapis/DapiServer.sol#L1035). Now that it's a median the check has been replaced with an equality check.

The timestamp equality check often fails because of an underlying issue; the child beacons on chain may not be up to date. Airseeker does a beaconset update check based on off-chain values (in the Airseeker cache), but the update function on the contract uses on-chain data, which may be stale. Therefore it is necessary to check that the underlying beacons on chain are fresh before attempting a parent update.

In the original update function, where we didn't care about the on-chain child data, this was fine... but now we update with only the beaconIds, relying on the on-chain data.

> I would definitely like to run this locally to be able to debug the code.

I'll pause work on this then and send you the config file on Keybase :)

acenolaza:
> It looks like the function that enabled updating a beacon set with signed data was removed from airnode-protocol-v1. In other words, in the current implementation that only way of updating a beacon set appears to be by first updating the constituents. AFAIK this is now the approach.
> 
> Assuming the above is correct, it then becomes a question of how and when those child data feeds should be updated; multicall, definitely, but do we update them with their existing-configured update triggers or do we perhaps do a more advanced check where we try to determine if updating a child would have a material effect on the parent?
> 
> Like maybe iterate through children, for every child check if the beaconset should be updated with all feeds excluding that child, if no, update it as part of the main update.

Oh I see what you mean. Thanks for adding more details and you are right. For updating a beaconSet it is now expected that most beacons are up-to-date but I'm still not sure who/how/when should those beacons be updated. I'm now thinking that this might be reason behind the decision to update the beaconSet beacons but I might have missed that discussion 🤔 

> In the original update function, where we didn't care about the on-chain child data, this was fine... but now we update with only the beaconIds, relying on the on-chain data.

👍🏻

> I'll pause work on this then and send you the config file on Keybase :)

 🙏🏻

bbenligiray:
The plan was to update the underlying Beacons and then the Beacon set in a single multicall, this was only discussed on Slack during the audit revisions. Let me know if we need to do a call about this.

aquarat:
Closed as changed integrated into the rate limiting fixes branch.

PR #393, opened 2023-05-17
216a Batch Beacon/Set follow-up fixes

aquarat:
One minor fix to the rate limiter code and another fix to the algorithm for updating child beacons.

### DIFF
--- a/src/calculations.ts
+++ b/src/calculations.ts
@@ -1,4 +1,4 @@
-import { ethers } from 'ethers';
+import { BigNumber, ethers } from 'ethers';
 import { HUNDRED_PERCENT } from './constants';
 
 export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, updatedValue: ethers.BigNumber) => {
@@ -21,10 +21,6 @@ export const calculateMedian = (arr: ethers.BigNumber[]) => {
   return arr.length % 2 !== 0 ? nums[mid] : nums[mid - 1].add(nums[mid]).div(2);
 };
 
-// TODO It isn't correct, fix
-// TODO: check Api3ServerV1 to determine if this is still correct
-// Calculate beacon set timestamp from beacon timestamps (https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/dapis/DapiServer.sol#L443)
-export const calculateBeaconSetTimestamp = (beaconSetBeaconTimestamps: string[]) => {
-  const accumulatedTimestamp = beaconSetBeaconTimestamps.reduce((total, next) => total + parseInt(next, 10), 0);
-  return Math.floor(accumulatedTimestamp / beaconSetBeaconTimestamps.length);
-};
+// https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/DataFeedServer.sol#L164
+export const calculateBeaconSetTimestamp = (beaconSetBeaconTimestamps: string[]) =>
+  calculateMedian(beaconSetBeaconTimestamps.map((ts) => BigNumber.from(ts)));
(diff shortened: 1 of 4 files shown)


src/validation.ts
```diff
@@ -398,6 +398,25 @@ export const signedDataSchema = z.object({
...
+  .merge(baseBeaconUpdateSchema)
+  .strict();
+
+export const beaconSetUpdateSchema = z
+  .object({
+    beaconSetId: config.evmIdSchema,
+  })
+  .merge(baseBeaconUpdateSchema)
+  .strict();
+
```
acenolaza: I renamed these to make update-data-feeds.ts more readable 😁 The are in lines [136-153](https://github.com/api3dao/airseeker/blob/c7a7fbb9f8098e4324c81b0d99e5fb52d30d4b71/src/validation.ts#L136-L153)

But if you added them back is because I probably missed changing the names somewhere 🤔

src/update-data-feeds.ts
```diff
@@ -488,10 +487,10 @@ export const updateBeaconSets = async (providerSponsorDataFeeds: ProviderSponsor
...
           }
 
-          // Get the beacon trigger from the config which has deviationThreshold and
-          // heartbeatInterval used in the condition checks down below
-          const beaconTrigger = beaconTriggers.find(
-            ({ beaconId }) => beaconId.toLowerCase() === beaconId.toLowerCase()
+          // Get the beaconSet trigger from the config which has deviationThreshold and
+          // heartbeatInterval used in the condition checks down below and use it against all child beacons
+          const beaconTrigger = beaconSetTriggers.find(
+            ({ beaconSetId }) => beaconSetId.toLowerCase() === beaconSetId.toLowerCase()
```
acenolaza: I was actually not sure about if we should use the beacon deviation/heartbeat or the beaconSet deviation/heartbeat during the condition check for each beacon. I decided to use the individual beacon trigger values just in case but I can see that it might become an issue if deviation for a beacon is set higher than the actual beaconSet.

So if we are just going to be using the beaconSet trigger values here then we can just remove these lines (490-501) and use  `beaconSetUpdateData.beaconSetTrigger` when calling `conditionCheck()` for each beacon (line 505).

src/calculations.ts
```diff
@@ -20,3 +20,7 @@ export const calculateMedian = (arr: ethers.BigNumber[]) => {
   });
   return arr.length % 2 !== 0 ? nums[mid] : nums[mid - 1].add(nums[mid]).div(2);
 };
+
+// https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/DataFeedServer.sol#L164
+export const calculateBeaconSetTimestamp = (beaconSetBeaconTimestamps: string[]) =>
+  calculateMedian(beaconSetBeaconTimestamps.map((ts) => BigNumber.from(ts)));
```
acenolaza: Is this here to replace lines 547-549?

acenolaza (commented):
Thanks for taking care of these issues 👏🏻

src/validation.ts
aquarat: Ah it's possible they're not needed - I'll check.
I'm not sure why but when I merged in main into your branch (which should have probably been done here instead) I got a whole bunch of build errors around missing types/variables. This was all an attempt to fix those errors.

src/update-data-feeds.ts
aquarat: Hey, for managed feeds we'll only ever update beaconsets and never individual beacons except as part of a beaconset update. Nodary will update individual beacons as part of their scheme, which may reduce our fees for managed feeds.

Thanks for the info, I've just made that change @ remove lines :)

src/calculations.ts
aquarat: Yes, and I see it actally isn't used 😝, I've removed it 💪

src/validation.ts
aquarat: Ah, I found the issue. I had a copy of one of the files that I used for experiments, it was uncommitted and it referenced the old names 🙈 .

acenolaza (approved):
🥇

PR #400, opened 2023-05-21
Filter empty sponsors

### LINKED ISSUE #309: Check sponsor wallet balances initially and omit the ones that are not funded
In the case that a sponsor wallet is not funded, Airseeker still goes through the entire flow until it attempts to make a transaction, which will fail. This is very wasteful for a bring-your-own-gas kind of scenario, where most sponsor wallets will be unfunded and you don't want to make unnecessary API calls for them. The suggested solution:
> Keep a `isFunded` flag for each sponsor wallet. When Airseeker encounters a "trigger" for a sponsor wallet, it first checks if it has already checked if it is funded. If it has and it was funded, continue. If it has and it wasn't funded, skip. If it hasn't, check if the sponsor wallet is funded, set the flag and proceed accordingly.

@bdrhn9 has implemented a solution that checks all sponsor wallets at the start. However, the suggested solution would also work well with [staggering](https://github.com/api3dao/airseeker/issues/308).

Note that this feature can/should be adapted to RRP in the single-Lambda Airnode.

bdrhn9:
Closes #309 

I implemented this feature differently than what was suggested in the related issue. The reason behind this choice is primarily to optimize memory usage and expedite the overall process. The suggested implementation would require to keep all non-funded sponsors in memory, necessitating an iteration over them to filter out the unfunded sponsors using the `isFunded` attribute in each update cycle. My implementation, in contrast, is quite straightforward. It identifies funded sponsors at the beginning of the program, updates the `config.triggers.dataFeedUpdates` state to include only those funded sponsors, and allows Airseeker to perform its tasks as usual. Tho, if suggested approach makes more sense to you, please let me know. I'll quickly switch over.

Additionally, to leverage Promise.any, I updated version of ECMAScript from 2020 to 2021.


acenolaza:
So if I understood this correctly:
1. Burak suggested checking the balance of the sponsor wallets and store them in memory when iterating over the triggers from config.
2. Bedirhan suggested checking the balance of all sponsor wallets before iterating over the triggers and filter out all triggers for which the sponsor wallet has a balance of zero.

I think both approaches are valid and at the end of the day, both require the same amount of RPC calls (which is what concerns me the most). I guess option #2 might be slightly easier to merge with currently being tested multicall branch. Also option #2 might make it easier to batch all rpc calls if that's ever supported by ethers.js (or by calling a contract that checks the balances of multiple EOA).

bbenligiray:
2 also has the advantage of having been tested for a while now

> option https://github.com/api3dao/airseeker/pull/2 might make it easier to batch all rpc calls if that's ever supported by ethers.js (or by calling a contract that checks the balances of multiple EOA)

Good point, Api3ServerV1 inherits ExtendedSelfMulticall to enable that. I created https://github.com/api3dao/airseeker/issues/406

bdrhn9:
So we are going with the solution 2? If so, this PR is surely ready to review. Thank you both for your comments :)

src/wallets.ts
```diff
@@ -34,6 +51,90 @@ export const initializeSponsorWallets = () => {
...
+  if (!goAnyResult.success) {
+    const message = `Failed to get balance for ${sponsorWalletAddress}. No provider was resolved. Error: ${goAnyResult.error.message}`;
+    logger.warn(message, logOptions);
+    return null;
+  }
+  const isEmpty = goAnyResult.data;
+  return { sponsorAddress, chainId, isEmpty };
+};
+
+export const filterEmptySponsors = async () => {
```
acenolaza, line 98: Just an idea but shouldn't we merge this function into [initializeSponsorWallets](https://github.com/api3dao/airseeker/blob/3555b1b3b6034afb9d06886e87458e9552fbadd8/src/wallets.ts#L33-L52) since we are already iterating over the sponsor list from triggers? This way we will also avoid deriving the private key for wallets that might not be going to be used.

bdrhn9: Ah, it'd be great but we need to use these private keys while deriving sponsor wallet addresses [here](https://github.com/api3dao/airseeker/blob/3555b1b3b6034afb9d06886e87458e9552fbadd8/src/wallets.ts#L58). So initializeSponsorWallets process is needed to be done before filtering. But we can also prune state.sponsorWalletsPrivateKey [here](https://github.com/api3dao/airseeker/blob/3555b1b3b6034afb9d06886e87458e9552fbadd8/src/wallets.ts#L128).

acenolaza (approved):
With the minor change suggested [here](https://github.com/api3dao/airseeker/pull/400#discussion_r1204383158) I think this LGTM 🔥

PR #401, opened 2023-05-21
Remove duplicate call for initialization

### LINKED ISSUE #399: Optimize initialization of update cycle
Both `updateBeacons` and `updateBeaconSets` functions currently call `initializeUpdateCycle` with almost identical parameters, except for `dataFeedType`. However, `dataFeedType` is only used in `logOptions`, which is not crucial and can be moved into the respective functions. This implementation has a disadvantage. For instance, if we have 8000 sponsors, each with only one beacon (as is the typical case), we would fetch 8000 transaction counts in `updateBeacons` and another 8000 transaction counts in `updateBeaconSets` unnecessarily, even though we don't have any beacon sets. To address this issue, I suggest the following modification:

1. Remove `initializeUpdateCycle` from both `updateBeacons` and `updateBeaconSets` functions. 
2. Remove parameter `dataFeedType` for `initializeUpdateCycle` which is used only for logOptions 
3. Move it just before where `updateBeacons` and `updateBeaconSets` functions are called, [here](https://github.com/api3dao/airseeker/blob/main/src/update-data-feeds.ts#L90)
4. Extend logOptions in `updateBeacons` and `updateBeaconSets` functions by adding `dataFeedType`

bdrhn9:
Closes #399

### DIFF
--- a/src/update-data-feeds.ts
+++ b/src/update-data-feeds.ts
@@ -1,4 +1,4 @@
-import { Api3ServerV1__factory as Api3ServerV1Factory } from '@api3/airnode-protocol-v1';
+import { Api3ServerV1, Api3ServerV1__factory as Api3ServerV1Factory } from '@api3/airnode-protocol-v1';
 import { go } from '@api3/promise-utils';
 import { getGasPrice } from '@api3/airnode-utilities';
 import { ethers } from 'ethers';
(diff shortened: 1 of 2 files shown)


bdrhn9:
The solution in the related issue that I suggested is problematic because it calls `initializeUpdateCycle` once as you can see [here](https://github.com/api3dao/airseeker/blob/fc6bb936a36a8e388244384683ef9309345eaac8/src/update-data-feeds.ts#L104). The problem here, if sponsor have both beacons and beacon sets and if sponsor updates at least one beacon, transaction counts is going to be deviated and upcoming beacon set updates will fail. So I ignored the first proposed solution and reverted the commit.

The last solution is straight-forward and not suffering from same problem.

src/update-data-feeds.ts
```diff
@@ -119,6 +118,15 @@ export const initializeUpdateCycle = async (
 
   logger.debug(`Initializing updates`, logOptions);
 
+  if (
+    (dataFeedType === DataFeedType.Beacon && beacons.length === 0) ||
+    (dataFeedType === DataFeedType.BeaconSet && beaconSets.length === 0)
```
acenolaza: Nit: since we are already importing lodash `_.isEmpty()` on this file we might as well use it:

```suggestion
    (dataFeedType === DataFeedType.Beacon && isEmpty(beacons)) ||
    (dataFeedType === DataFeedType.BeaconSet && isEmpty(beaconSets))
```

acenolaza (approved):
LGTM 👍🏻

PR #409, opened 2023-05-26
[Draft] Airseeker Telemetry

aquarat:
[Draft] I have to merge in Santiago's recent fixes.
This adds logging and alerting to Airseeker.
The intention is for this to not be merged into the main app but rather be a separate branch indefinitely.

The app logs current off-chain values of feeds to a database and alerts based on Airseeker update conditions with a multiplier applied.

The idea is that this database can run in a separate "schema" to our main DB, possibly on the same server (staging).


aquarat:
Closed as changes rebased on a different branch.

PR #413, opened 2023-05-30
Alerting and logging

aquarat:
This adds logging and alerting to Airseeker.
The intention is for this to not be merged into the main app but rather be a separate branch indefinitely.

The app logs current off-chain values of feeds to a database and alerts based on Airseeker update conditions with a multiplier applied.

The idea is that this database can run in a separate "schema" to our main DB, possibly on the same server (staging).

The general approach is:
Based on config (quoted below), do the assessment/checkCondition phase for all beacons, beaconSets and constituents of beaconSets. For every beacon and beaconSet, check the condition with a multiplier applied and log the value to the database. Jump out of the scope before getting to the chain txn creation operations.

The mnemonic used in testing is a dummy mnemonic (eg. "test test test... junk").

Sponsor wallets are not queried in `monitorOnly` mode.

```json
{
  "...": "...",
  "monitoring": {
    "monitorOnly": true,
    "deviationMultiplier": 1.1,
    "heartbeatMultiplier": 1.1,
    "opsGenieApiKey": "f152270b-7ca5-476c-b050-227bb45238d1",
    "heartbeatId": "airseeker-telemetry"
  },
  "...": "...",
}
```

Closes https://github.com/api3dao/telemetry/issues/59


prisma/schema.prisma
```diff
@@ -0,0 +1,35 @@
...
+  type       DataFeedType
+  @@index([dataFeedId])
+}
+
+model DeviationValue {
+  id         String   @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  when       DateTime @default(now())
+  dataFeedId String
+  chainId    String
+  deviation  Float
```
acenolaza, line 33: I remember we used `Decimal   @db.Decimal(5, 2)`  for deviation in operations-database. Might be useful here too.

src/alerting.ts
```diff
@@ -0,0 +1,154 @@
...
+      },
+    }),
+    prisma?.deviationValue.create({
+      data: {
+        dataFeedId,
+        deviation: parseFloat(calculateUpdateInPercentage(onChainValue, offChainValue).toString()) / HUNDRED_PERCENT,
+        chainId,
+      },
+    }),
+  ]);
```
acenolaza: If this 2 inserts need to happen atomically then we might need to wrap them with a transaction 🤔

src/alerting.ts
```diff
@@ -0,0 +1,154 @@
...
+    await Promise.allSettled([
+      limitedCloseOpsGenieAlertWithAlias(
+        generateOpsGenieAlias(`${UpdateStatus.ON_CHAIN_TIMESTAMP_OLDER_THAN_HEARTBEAT_MESSAGE}${dataFeedId}${chainId}`),
+        opsGenieConfig
+      ),
+      limitedCloseOpsGenieAlertWithAlias(
+        generateOpsGenieAlias(`${UpdateStatus.DEVIATION_THRESHOLD_REACHED_MESSAGE}${dataFeedId}${chainId}`),
+        opsGenieConfig
+      ),
+    ]);
```
acenolaza: This has caught me scratching my head recently, in operations-database we usually [send an alert and close it right after we send it](https://github.com/api3dao/operations-database/blob/main/src/workers/beacon-events.ts#L323-L333). Here you are only closing the alert while deviation and heartbeat are not reached. Should we start doing the same in operations-database?

src/check-condition.ts
```diff
@@ -1,38 +1,45 @@
...
+}
 
 export const checkConditions = (
   onChainDataValue: ethers.BigNumber,
   onChainDataTimestamp: number,
   fulfillmentDataTimestamp: number,
   trigger: Pick<BeaconTrigger | BeaconSetTrigger, 'deviationThreshold' | 'heartbeatInterval'>,
   apiValue: ethers.BigNumber
-): LogsData<boolean> => {
+): [PendingLog[], boolean, UpdateStatus] => {
```
acenolaza: Nit: here you could keep the LogsData return type but instead of just boolean you could use an object. Something like this:

```
interface CheckConditionsResult {
  result: boolean;
  updateStatus: UpdateStatus;
}

export const checkConditions = (
  onChainDataValue: ethers.BigNumber,
  onChainDataTimestamp: number,
  fulfillmentDataTimestamp: number,
  trigger: Pick<BeaconTrigger | BeaconSetTrigger, 'deviationThreshold' | 'heartbeatInterval'>,
  apiValue: ethers.BigNumber
): LogsData<CheckConditionsResult> => {
  // Check that fulfillment data is newer than on chain data
  const isFulfillmentDataFresh = checkFulfillmentDataTimestamp(onChainDataTimestamp, fulfillmentDataTimestamp);
  if (!isFulfillmentDataFresh) {
    const log = logger.pend('WARN', UpdateStatus.FULFILLMENT_DATA_OLDER_THAN_ON_CHAIN_MESSAGE);
    return [[log], { result: false, updateStatus: UpdateStatus.FULFILLMENT_DATA_OLDER_THAN_ON_CHAIN_MESSAGE }];
  }

}
```

src/validation.ts
```diff
@@ -366,6 +366,16 @@ const validateDataFeedUpdatesReferences: SuperRefinement<{
 export const configSchema = z
   .object({
     airseekerWalletMnemonic: z.string(),
+    monitoring: z
+      .object({
+        monitorOnly: z.boolean(),
+        deviationMultiplier: z.number(),
+        heartbeatMultiplier: z.number(),
+        opsGenieApiKey: z.string(),
```
acenolaza:
```suggestion
        opsGenieApiKey: z.string().uuid(),
```

src/validation.ts
```diff
@@ -366,6 +366,16 @@ const validateDataFeedUpdatesReferences: SuperRefinement<{
 export const configSchema = z
   .object({
     airseekerWalletMnemonic: z.string(),
+    monitoring: z
+      .object({
+        monitorOnly: z.boolean(),
+        deviationMultiplier: z.number(),
+        heartbeatMultiplier: z.number(),
```
acenolaza: Can these be negative or zero? Otherwise:
```suggestion
        deviationMultiplier: z.number().positive(),
        heartbeatMultiplier: z.number().positive(),
```

src/update-data-feeds.ts
```diff
@@ -273,6 +277,25 @@ export const updateBeacons = async (providerSponsorDataFeeds: ProviderSponsorDat
         beaconReturndata
       );
 
+      if (monitorOnly) {
+        await Promise.allSettled([
+          checkAndReport(
+            'Beacon',
+            prisma!,
```
acenolaza: Just curious, why pass the client as param instead of importing it in alerting.ts?

acenolaza (approved):
A few minor comments but overall LGTM 🚀

src/alerting.ts
aquarat: Good note, but they don't need to happen atomically/they're independent. The use of `allSettled` was just to elegantly handle errors. This is a fire-and-forget action. :)

src/alerting.ts
aquarat: That link might be broken 😅 
But we're not supposed to close an alert immediately after creating it. How it works is, if something goes wrong (the tank pressure is too high), create an alert. If the tank pressure drops again, close that same alert (because the triggering condition no longer applies).

So you'll often find something in code along the lines of
```typescript
myFunction = () => {
try {
    aFunctionThatThrows();
} catch (e) {
    sendToOpsGenie({
        message: "something",
        priority: "P2",
        alias: "something"
    });
    return;
}

closeOpsGenieAlertWithAlias("something");
};
```
The `return` is important: one will fire if there's an error and the other will be executed if the error doesn't occur.

A practical example from the file you linked:
```typescript
        const chainConfig = config.chains[chainData.id];
        if (!chainConfig) {
          await limitedSendToOpsGenieLowLevel(
            {
              priority: 'P3',
              alias: `missing-chain-config-error`,
              message: `No chain config found for chain with id (${chainData.id}), please check config.json.`,
            },
            config.opsGenieConfig
          );
          throw new Error(`No chain config found for ${chainData.name}, please check config.json.`);
        }
        await limitedCloseOpsGenieAlertWithAlias('missing-chain-config-error', config.opsGenieConfig);
```
Here the closeAlert call won't fire if the alert in the if above is executed.

In the PR code this comment is attached to the close alert calls will only run if no other conditions are satisfied (it's a tree of if statements, which I dislike, but that's life hah). In other words, if no alert scenario occurs, it'll close the alerts. Due to the construction of the aliases it will only close alerts related to that exact feed on that specific chain.

It's quite cumbersome, I have wondered if there's a better way of handling this, but I haven't been able to come up with anything better.

src/validation.ts
aquarat: oh nice 🙏

src/update-data-feeds.ts
aquarat: You're right, this is actually a leftover from a different approach. I'll remove it 😅

src/check-condition.ts
aquarat: I agree that's a lot cleaner. The reason I did it this way was to retain the existing code's interface to make future merges easier... but I think it's beyond that point now haha. I'll adopt this approach 👍

prisma/schema.prisma
aquarat: I just remembered there was a big cloud of issues and questions around the use of Decimal and I wanted to avoid that.

Float in postgres is the equivalent of Postgres's "double-precision" aka many languages' `float64`. In terms of storage and DB performance, float is much more efficient than Decimal and in terms of our code it maps 1:1 to JS's float numbers without the use a third-party library. I learnt from the last Telemetry logging app that keeping the data types simple made for easier use later on in tools like Grafana. Deviation here is a low accuracy number, we mostly want to use it for graphing and alerting. Same thing with the on-chain values - we don't really care about the precision beyond a certain point, we just need to be able to make pretty charge :)

aquarat:
Thanks again for the review Santiago, much appreciated :)

src/alerting.ts
acenolaza: Thanks for the detailed answer with examples and everything 👏🏻 I'm now thinking i might have missed a `return;` in the last worker I added 🤦🏻‍♂️

PR #432, opened 2023-06-29
Airseeker monitoring changes to accommodate Grafana

aquarat:
In preparation for the 14th of August we required additional monitoring and a lot of changes to the existing monitoring system/apps.

Generally the approach was to try and make the logged data as easy as possible to query - so this PR tries to add that extra data.

There's a lot of mutating going on, that can't be helped.

I suggest the reviewer not worry about the migration files and instead go straight to alerting.ts and /src/ changes.

This branch is being actively used.

And apologies for the big PR.


src/alerting.ts
```diff
@@ -159,3 +615,155 @@ export const checkAndReport = async (
     ]);
   }
 };
+
+// TODO this needs to come out of the DB or some other static resource
+export const DB_CHAINS = [
```
vponline, line 620: Could this be replaced with the CHAINS export from the [chains](https://github.com/api3dao/chains/blob/main/src/generated/chains.ts) repo? The `alias` field there corresponds to the `ChainInfrastructure.name` field.

src/alerting.test.ts
```diff
@@ -29,11 +71,14 @@ describe('alerting', () => {
     // eslint-disable-next-line no-console
     mockLimitedSendToOpsGenieLowLevel.mockImplementation(console.log);
     setOpsGenieHandlers(mockLimitedCloseOpsGenieAlertWithAlias, mockLimitedSendToOpsGenieLowLevel);
```
vponline, line 73: It's also possible to mock the operations-utilities package for these handlers like [here](https://github.com/api3dao/operations-database/blob/main/src/api/dapi-prices.test.ts#L4), but I don't mind the current way either 😄

vponline (approved):
This was much easier to understand than the old Telemetry 😄. Hopefully, maintaining this will be easier as well

src/alerting.ts
aquarat: Just saw this, I tried that originally but the aliases didn't match at the time. I'll try again. If they still don't match we may have to formally change them in the DBs. Warren also recently complained about the mismatch.

aquarat: The aliases don't match, I'll paste the mismatch here so we have something to refer to in future. Given the priorities we probably shouldn't fix this 🤷 

```
Chain alias mismatch for ID          1 | Package alias: ethereum                        DB alias: mainnet                       
Chain alias mismatch for ID   11155111 | Package alias: ethereum-sepolia-testnet        DB alias: sepolia                       
Chain alias mismatch for ID       1442 | Package alias: polygon-zkevm-goerli-testnet    DB alias: polygon-zkevm-testnet         
Chain alias mismatch for ID       2001 | Package alias: milkomeda-c1                    DB alias: milkomeda                     
Chain alias mismatch for ID     200101 | Package alias: milkomeda-c1-testnet            DB alias: milkomeda-testnet             
Chain alias mismatch for ID        420 | Package alias: optimism-goerli-testnet         DB alias: optimism-testnet              
Chain alias mismatch for ID     421613 | Package alias: arbitrum-goerli-testnet         DB alias: arbitrum-testnet              
Chain alias mismatch for ID          5 | Package alias: ethereum-goerli-testnet         DB alias: goerli                        
Chain alias mismatch for ID        599 | Package alias: metis-goerli-testnet            DB alias: metis-testnet     
```

PR #461, opened 2023-08-18
Increase RPC and provider timeout constants

### LINKED ISSUE #460: Increase Provider and RPC timeouts to 15 seconds (currently 5 seconds)
https://github.com/api3dao/airseeker/blob/main/src/constants.ts#L3

aquarat:
Closes #460

I'm running this locally with the production config and so far it looks really good; very few gateway failures.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -3,7 +3,7 @@ import { ethers } from 'ethers';
 export const GATEWAY_TIMEOUT_MS = 15_000;
 export const PROVIDER_TIMEOUT_MS = 15_000;
 export const RANDOM_BACKOFF_MIN_MS = 0;
-export const RANDOM_BACKOFF_MAX_MS = 15_000;
+export const RANDOM_BACKOFF_MAX_MS = 2_500;
 export const PRIORITY_FEE_IN_WEI = 3_120_000_000;
 // The Base Fee to Max Fee multiplier
 export const BASE_FEE_MULTIPLIER = 2;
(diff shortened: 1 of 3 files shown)


src/constants.ts
```diff
@@ -1,9 +1,9 @@
 import { ethers } from 'ethers';
 
-export const GATEWAY_TIMEOUT_MS = 5_000;
-export const PROVIDER_TIMEOUT_MS = 5_000;
+export const GATEWAY_TIMEOUT_MS = 15_000;
+export const PROVIDER_TIMEOUT_MS = 15_000;
```
acenolaza, line 4: So not only gateways but RPCs are also timing out with 5secs?

acenolaza (approved):
LGTM 👍🏻 Merge it whenever you feel comfortable that it doesn't generate any side effects 💪🏻

src/constants.ts
acenolaza: nvm, I just saw this comment 😄 https://api3workspace.slack.com/archives/C05MXJF7831/p1692352864842519?thread_ts=1692352070.363849&cid=C05MXJF7831

src/constants.ts
```diff
@@ -1,9 +1,9 @@
...
 export const RANDOM_BACKOFF_MIN_MS = 0;
-export const RANDOM_BACKOFF_MAX_MS = 2_500;
+export const RANDOM_BACKOFF_MAX_MS = 15_000;
```
bdrhn9: I think that 15 secs backoff is too large. I'd leave it as is. Other than that it LGTM 🔥

bdrhn9 (approved):

Issue #447, opened 2023-07-17
Support undefined fulfillment gas limit


bbenligiray:
Airnode v0.12 will make `fulfillmentGasLimit` in `chains` an optional fields. Airnode needs to [jump through a lot of hoops](https://github.com/api3dao/airnode/issues/1589) to support that. ~Airseeker just needs to omit the `gasLimit` field in the transaction, which will tell ethers to make an `estimateGas()` call beforehand.~ (see below) This should be useful for (multicall-based) Beacon set updates, which will differ in the required gas limit.

This also makes chain integrations easier and safer, as the required gas limit floats in some ~rollups~L2s, making hardcoding a gas limit a risky strategy.

bbenligiray:
Based on https://github.com/api3dao/manager-multisig/pull/214#issuecomment-1663995357 the original description is wrong. If you omit the `gasLimit` of a `tryMulticall` transaction, the estimated gas limit will be wrong.

PR #472, opened 2023-09-08
446 improve sponsor wallet filtering

### LINKED ISSUE #446: Improve sponsor wallet filtering
Currently, sponsor wallet balances are fetched at the beginning of the coordinator, and the sponsor wallets that have exactly zero balance are filtered out to reduce the Airseeker load. Self-funded sponsor wallets update until the funds run out, but the funds leave some dust instead of running out completely, and the filtering strategy isn't suited to address this.

Proposed method:
- At the start of the coordinator, generate a random wallet (to use as `airnode`), and call `estimateGas()` with it to estimate the gas of initializing a Beacon
- As the second step, fetch the gas price as described by the gas price strategy
- Filter the sponsor wallets whose balances are lower than the multiplication of the above two values

acenolaza:
Closes #446

### DIFF
--- a/src/wallets.test.ts
+++ b/src/wallets.test.ts
@@ -1,9 +1,3 @@
-import {
-  ChainOptions,
-  ConstantGasPriceStrategy,
-  GasPriceOracleConfig,
-  ProviderRecommendedGasPriceStrategy,
-} from '@api3/airnode-node';
 import { logger } from '@api3/airnode-utilities';
 import { ethers } from 'ethers';
 import { RateLimitedProvider } from './providers';
@@ -66,27 +60,6 @@ const config = {
     },
   },
 } as unknown as Config;
-
-const providerRecommendedGasPriceStrategy: ProviderRecommendedGasPriceStrategy = {
-  gasPriceStrategy: 'providerRecommendedGasPrice',
-  recommendedGasPriceMultiplier: 1.2,
-};
-const constantGasPriceStrategy: ConstantGasPriceStrategy = {
-  gasPriceStrategy: 'constantGasPrice',
-  gasPrice: {
-    value: 10,
-    unit: 'gwei',
-  },
-};
-const defaultGasPriceOracleOptions: GasPriceOracleConfig = [
-  providerRecommendedGasPriceStrategy,
-  constantGasPriceStrategy,
-];
-const fulfillmentGasLimit = 500_000;
-const defaultChainOptions: ChainOptions = {
-  gasPriceOracle: defaultGasPriceOracleOptions,
-  fulfillmentGasLimit,
-};
 
 beforeEach(() => {
   state.initializeState(config);
@@ -151,6 +124,11 @@ describe('hasEnoughBalance', () => {
         getBlock: jest.fn().mockResolvedValue({
           timestamp: Math.floor(Date.now() / 1000),
         }),
+        network: {
+          chainId: '1',
+          name: 'mainnet',
+        },
+        getGasPrice: jest.fn().mockResolvedValue(ethers.utils.parseUnits('10', 'gwei')),
       },
       getBalance: jest.fn().mockResolvedValue(ethers.utils.parseEther('1')),
     };
@@ -166,13 +144,7 @@ describe('hasEnoughBalance', () => {
       },
     };
 
-    const result = await hasEnoughBalance(
-      defaultChainOptions,
-      sponsorWallet as any,
-      airnode as any,
-      api3ServerV1 as any,
-      logOptions
-    );
+    const result = await hasEnoughBalance(sponsorWallet as any, airnode as any, api3ServerV1 as any, logOptions);
 
     expect(result).toBeTruthy();
   });
@@ -183,6 +155,11 @@ describe('hasEnoughBalance', () => {
         getBlock: jest.fn().mockResolvedValue({
           timestamp: Math.floor(Date.now() / 1000),
         }),
+        network: {
+          chainId: '1',
+          name: 'mainnet',
+        },
+        getGasPrice: jest.fn().mockResolvedValue(ethers.utils.parseUnits('10', 'gwei')),
       },
       getBalance: jest.fn().mockResolvedValue(ethers.utils.parseUnits('1', 'wei')),
     };
@@ -198,13 +175,7 @@ describe('hasEnoughBalance', () => {
       },
     };
 
-    const result = await hasEnoughBalance(
-      defaultChainOptions,
-      sponsorWallet as any,
-      airnode as any,
-      api3ServerV1 as any,
-      logOptions
-    );
+    const result = await hasEnoughBalance(sponsorWallet as any, airnode as any, api3ServerV1 as any, logOptions);
 
     expect(result).toBeFalsy();
   });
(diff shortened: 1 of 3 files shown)


src/wallets.ts
```diff
@@ -51,48 +54,131 @@ export const initializeSponsorWallets = () => {
...
+    logger.error('Failed to get the block', goGetBlock.error, logOptions);
+    throw new Error(goGetBlock.error.message);
+  }
+  // Check if last block is not older than 5 min
+  const lastBlock = goGetBlock.data;
+  const fiveMinutesAgo = Math.floor(Date.now() / 1000) - 60 * 5;
+  if (lastBlock.timestamp < fiveMinutesAgo) {
+    logger.warn('Last block is older than 5 minutes. SponsorWallets will be skipped for this chain', logOptions);
+    throw new Error(`Chain has not produced a new block in the last 5 minutes`);
+  }
```
acenolaza: I'm not sure about this. I added it after a comment @bbenligiray made on slack about doing this check on operations workers and collectors.

src/wallets.test.ts
```diff
@@ -77,152 +128,215 @@ describe('retrieveSponsorWalletAddress', () => {
...
-
-    const sponsorWalletAddress = 'sponsorWalletAddress';
-    const expectedErrorMessage = 'RPC Error while retrieving balance';
-
-    await expect(wallets.isBalanceZero(rpcProvider, sponsorWalletAddress)).rejects.toThrow(expectedErrorMessage);
-
-    expect(rpcProvider.getBalance).toHaveBeenCalledTimes(2);
-    expect(rpcProvider.getBalance).toHaveBeenCalledWith(sponsorWalletAddress);
-  });
+  // TODO: add more tests
```
acenolaza: I need to add more tests for the unhappy paths

src/wallets.ts
```diff
@@ -51,73 +54,166 @@ export const initializeSponsorWallets = () => {
...
+  sponsorWallet: ethers.Wallet,
+  airnode: ethers.Wallet,
+  api3ServerV1: ethers.Contract,
+  logOptions: LogOptionsOverride
 ): Promise<boolean> => {
-  const goResult = await go(() => rpcProvider.getBalance(sponsorWalletAddress), { retries: 1 });
-  if (!goResult.success) {
-    throw new Error(goResult.error.message);
+  // Fetch current block number
+  const goGetBlock = await go(() => sponsorWallet.provider.getBlock('latest'), { retries: 1 });
```
bdrhn9: I wouldn't prefer to get this done here because for example Nodary has 417 wallets for each chain currently. It means it calls RPC 417 times to get same block only for one chain. I'm thinking total ~12.5k wallets when all chains included, it probably won't work for Nodary. That being said, I think we shouldn't add this block validation thing into Airseeker because there is no database to disturb with malfunctioning RPC. Also most of L2 chains don't produce block if there is no transaction, that's why I wouldn't prefer to add extra assumption (there will be always a transaction).

src/wallets.ts
```diff
@@ -51,73 +54,166 @@ export const initializeSponsorWallets = () => {
...
+    throw new Error(goEstimateGas.error.message);
+  }
+  const estimatedGas = goEstimateGas.data;
+
+  logger.info(`Current balance: ${ethers.utils.formatEther(balance)} ether`, logOptions);
+  logger.info(`Current gas price: ${ethers.utils.formatUnits(gasPrice, 'gwei')} gwei`, logOptions);
+  logger.info(`Estimated gas needed: ${estimatedGas.toString()}`, logOptions);
+
+  // Check if the current balance is enough to cover the estimated gas cost
+  return balance.gte(estimatedGas.mul(gasPrice));
```
bdrhn9, line 154: For mantle (and alike legacy Optimism forks), when gas price is fetched by `provider.getGasPrice()`, it returns only L2 gas price. So to check enough fund exists on a wallet, [rollup_gasPrices](https://docs.mantle.xyz/network/for-devs/resources-and-tooling/mantle-node-api#rollup_gasprices) needs to be utilized. As a boilerplate, I'm using following code piece in my monitoring system:
```typescript
export const getMantleGasPrice = async (provider: ethers.providers.JsonRpcProvider) => {
  const goRes = await go(() => provider.send('rollup_gasPrices', []));
  if (!goRes.success) throw goRes.error;

  const l1GasPrice = ethers.BigNumber.from(goRes.data.l1GasPrice);
  const l2GasPrice = ethers.BigNumber.from(goRes.data.l2GasPrice);

  return l1GasPrice.add(l2GasPrice);
};

export const getGasPrice = async (chainId: string, provider: ethers.providers.JsonRpcProvider) => {
  switch (chainId) {
    case '5000':
    case '5001':
      return getMantleGasPrice(provider);
    default:
      return provider.getGasPrice();
  }
};
```

src/wallets.ts
```diff
@@ -51,73 +54,166 @@ export const initializeSponsorWallets = () => {
...
+    log.error ? logger.error(log.error.message, null, logOptions) : logger.debug(log.message, logOptions)
+  );
+  if (!gasTarget) {
+    throw new Error('Failed to get gas price');
   }
-  return goResult.data.isZero();
+  const gasPrice = gasTarget.type === 2 ? gasTarget.maxFeePerGas : gasTarget.gasPrice;
+
+  // Estimate the units of gas required for updating a single dummy beacon with signed data
+  const goEstimateGas = await go(
```
bdrhn9: Why we are estimating gas instead of using `fulfillmentGasLimit` defined in airseeker configuration? As you know, because Nodary contains a large number of sponsors, I am concerned about making any additional API calls and calculations.

src/wallets.ts
acenolaza: All very valid points 👏🏻 I'll just remove this validation altogether. Thanks for the detailed explanation

src/wallets.ts
acenolaza: > Why we are estimating gas instead of using `fulfillmentGasLimit` defined in airseeker configuration? As you know, because Nodary contains a large number of sponsors, I am concerned about making any additional API calls and calculations.

My guess is because `fulfillmentGasLimit` is going to be refactored into an optional field. Also, `fulfillmentGasLimit` might be way larger than a single beacon update gas usage. I do like the suggestion tho because it reduces the number of RPC calls needed. @bbenligiray thoughts?

bdrhn9 (commented):
I'm okay with latest changes that removes block validation. I'm waiting comment for my `fulfillmentGasLimit` comment above. Addition to that, mantle-specific solution needs to be implemented to work on that network.

src/wallets.ts
```diff
@@ -51,73 +54,166 @@ export const initializeSponsorWallets = () => {
...
+    async () => {
+      const dummyBeaconTemplateId = ethers.utils.hexlify(ethers.utils.randomBytes(32));
+      const dummyBeaconTimestamp = Math.floor(Date.now() / 1000);
+      const randomBytes = ethers.utils.randomBytes(Math.floor(Math.random() * 27) + 1);
+      const dummyBeaconData = ethers.utils.defaultAbiCoder.encode(
+        ['int224'],
+        // Any radom number that fits into an int224
+        [ethers.BigNumber.from(randomBytes)]
+      );
+      const dummyBeaconSignature = await airnode.signMessage(
```
bbenligiray: Don't sign with `airnode`

acenolaza: I forgot that I was already using a [dummy](https://github.com/api3dao/airseeker/pull/472/files?show-viewed-files=true&file-filters%5B%5D=#diff-ac6a312c9e24133a7c95de8d3425130b31ffb3f35d99201839b67c76fac94770R208) ethers.Wallet. The fact that I called the function argument `airnode` caused the confusion.

src/wallets.ts
```diff
@@ -61,8 +60,27 @@ export const retrieveSponsorWallet = (sponsorAddress: string): ethers.Wallet =>
...
+
+  const l1GasPrice = ethers.BigNumber.from(goRes.data.l1GasPrice);
+  const l2GasPrice = ethers.BigNumber.from(goRes.data.l2GasPrice);
+
+  return l1GasPrice.add(l2GasPrice);
+};
+
+const getGasPrice = async (provider: RateLimitedProvider) => {
+  switch (provider.network.name) {
+    case 'mantle':
```
bdrhn9: I didn't get where ethers fetching network name 😄 To make it more convenient I'd rather to use chain ids.

src/wallets.ts
bdrhn9: I understood from yesterday meeting that we will stick with `fulfillmentGasLimit`.

src/wallets.ts
acenolaza: The `chainName` is being set [when instantiating](https://github.com/api3dao/airseeker/blob/446-improve-sponsor-wallet-filtering/src/providers.ts#L92) a new RateLimitedProvider object which is just a custom `ethers.providers.StaticJsonProvider` class. I thought the `chainName` was easier to read than the `chainId`. I can switch to `chainId` if you still prefer but I don't really see an issue with using the `chainName`. Am I missing something? 😅

src/wallets.ts
acenolaza: I understood that I should just keep estimating but I just needed to make sure the signature was done using a random wallet and not actual airnode wallet (this was just due to a [confusing argument name](https://github.com/api3dao/airseeker/pull/472/commits/1cdff8b677d3b1560e290ca2c88056fc8bf5a9ae)). I also think Burak wanted to keep the estimation code to then maybe re-use it when implementing [this](https://github.com/api3dao/airseeker/issues/447). WDYT?

acenolaza: BTW shouldn't we also update the [airnode-utils.fetchProviderRecommendedGasPrice](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts#L66) strategy to handle mantle as well 🤔

src/wallets.ts
bdrhn9: I'm a bit paranoiac and I'm trusting number much rather than texts :-D So I'd prefer to use them however I'm always respect for your preference.

src/wallets.ts
bdrhn9: > BTW shouldn't we also update the [airnode-utils.fetchProviderRecommendedGasPrice](https://github.com/api3dao/airnode/blob/master/packages/airnode-utilities/src/evm/gas-prices/gas-oracle.ts#L66) strategy to handle mantle as well 🤔

No, while sending transaction knowing L2 gas price is enough for this chain.

bdrhn9: I talked with @bbenligiray , he have better suggestion that checking whether chain has `fulfillmentGasLimit` or not (for the upcoming making it optional scenario), if it has, use this gas limit otherwise estimate gas unit. It will be also compatible with upcoming mentioned scenario.

acenolaza: Ok. I'll then use the `fulfillmentGasLimit` here and when actually making the update tx if it exist. Otherwise, call `estimateGas * gasPrice` 👌🏻 (on both places, wallet filtering and update txs)

(With the assumption that Nodary's airseeker.json is not using `fulfillmentGasLimit` of `2000000` like the [managed data feed Airseeker is using](https://github.com/api3dao/operations-airseeker-config/blob/main/airseeker.json#L453)).

bdrhn9 (approved):
LGTM 🔥

PR #481, opened 2023-09-22
Support undefined fulfillment gas limit

### LINKED ISSUE #447: Support undefined fulfillment gas limit

acenolaza:
Closes #447 

I'm basically doing the following:

1. check if `fulfillmentGasLimit` exist in the config
2. if it doesn't then estimate the gas limit
2.1. estimate based on Api3ServerV1.multicall (tryMulticall estimate is usually wrong). If successful then add an extra since tryMulticall is a bit more expensive
2.2. if previous failed, estimate based on a single beacon update using dummy data (for beacon set update we add up the estimate of updating all beacon plus the beaconSet)
2.3. if both previous failed then just return a hardcoded gas limit  of 2_000_000

This is still a draft because I need to fix/add some tests and do some minor cleanup/refactor

### DIFF
--- a/src/update-data-feeds.ts
+++ b/src/update-data-feeds.ts
@@ -344,50 +344,51 @@ export const updateBeacons = async (providerSponsorDataFeeds: ProviderSponsorDat
         logOptions
       );
 
-      const tx = await go(
-        updateBatch.length === 1
-          ? () =>
-              contract
-                .connect(sponsorWallet)
-                .updateBeaconWithSignedData(
-                  beaconUpdates[0].beacon.airnode,
-                  beaconUpdates[0].beacon.templateId,
-                  beaconUpdates[0].newBeaconResponse.timestamp,
-                  beaconUpdates[0].newBeaconResponse.encodedValue,
-                  beaconUpdates[0].newBeaconResponse.signature,
-                  { nonce, ...gasTarget }
-                )
-          : async () => {
-              const calldatas = updateBatch.map((beaconUpdateData) =>
-                contract.interface.encodeFunctionData('updateBeaconWithSignedData', [
-                  beaconUpdateData.beacon.airnode,
-                  beaconUpdateData.beacon.templateId,
-                  beaconUpdateData.newBeaconResponse.timestamp,
-                  beaconUpdateData.newBeaconResponse.encodedValue,
-                  beaconUpdateData.newBeaconResponse.signature,
-                ])
-              );
+      let updateFn;
+      if (updateBatch.length === 1) {
+        updateFn = () =>
+          contract
+            .connect(sponsorWallet)
+            .updateBeaconWithSignedData(
+              beaconUpdates[0].beacon.airnode,
+              beaconUpdates[0].beacon.templateId,
+              beaconUpdates[0].newBeaconResponse.timestamp,
+              beaconUpdates[0].newBeaconResponse.encodedValue,
+              beaconUpdates[0].newBeaconResponse.signature,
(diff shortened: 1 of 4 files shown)


src/update-data-feeds.ts
```diff
@@ -168,6 +177,45 @@ export const initializeUpdateCycle = async (
...
+const estimateBeaconMulticallGasLimit = async (
+  contract: Api3ServerV1,
+  calldatas: string[],
+  logOptions: LogOptionsOverride
+) => {
+  const estimateGasMulticall = await go(() => contract.estimateGas.multicall(calldatas), {
+    retries: 1,
+  });
+  if (estimateGasMulticall.success) {
+    // Adding a extra 10% because multicall consumes less gas than tryMulticall
```
bbenligiray, line 189: I came here to suggest adding a % overhead :+1:

acenolaza: The only thing that worries me now is that these changes should be an improvement over the `fulfillmentGasLimit` magic number but I'm now adding a different magic number which might even be configurable by chain 😅 
On tests against the local hardhat network even 1% was enough but I added 10% just to be more confident it will be enough on all chains.

bbenligiray: Yeah I very much doubt this will be a problem

acenolaza:
@bdrhn9 @aquarat friendly reminder to look at this whenever you guys have some time 🙏🏻 Thanks

src/utils.ts
```diff
@@ -10,3 +11,23 @@ export const calculateTimeout = (startTime: number, totalTimeout: number) => tot
...
   delay: { type: 'random' as const, minDelayMs: RANDOM_BACKOFF_MIN_MS, maxDelayMs: RANDOM_BACKOFF_MAX_MS },
 });
+
+export const createDummyBeaconUpdateData = async (dummyAirnode: ethers.Wallet = ethers.Wallet.createRandom()) => {
+  const dummyBeaconTemplateId = ethers.utils.hexlify(ethers.utils.randomBytes(32));
+  const dummyBeaconTimestamp = Math.floor(Date.now() / 1000);
+  const randomBytes = ethers.utils.randomBytes(Math.floor(Math.random() * 27) + 1);
+  const dummyBeaconData = ethers.utils.defaultAbiCoder.encode(
+    ['int224'],
+    // Any radom number that fits into an int224
```
bdrhn9:
```suggestion
    // Any random number that fits into an int224
```
Nit: typo in comment

bdrhn9 (approved):
It's well-designed and LGTM 🔥. Because I was planning to take vacation, I'd like to focus on tasks on`nodaryio` and `reblokio` organizations. So sorry for really late review.

acenolaza:
> It's well-designed and LGTM 🔥. Because I was planning to take vacation, I'd like to focus on tasks on`nodaryio` and `reblokio` organizations. So sorry for really late review.

No worries and thanks for checking this PR out @bdrhn9 🙏🏻

PR #489, opened 2023-10-18
Add additional info to alerts | Batch DB writes

aquarat:
This PR is currently being tested live.

This PR does two main things:
- Batches DB writes to improve performance/reduce DB load
- Includes extra info in OpsGenie alerts related to dead gateways (how many out of how many gateways are dead and whether those gateways are AWS/GCP). This will make it easier to figure out if a dead gateway is a big deal or not.
- Updates the hard-coded chains :( 

Closes https://github.com/api3dao/tasks/issues/219
Closes https://github.com/api3dao/tasks/issues/199

### DIFF
--- a/src/alerting.ts
+++ b/src/alerting.ts
@@ -60,8 +60,11 @@ const writeRecords = async () => {
     )
   );
 
-  // eslint-disable-next-line no-console
-  console.error(results.filter((result) => result.status === 'rejected'));
+  const dbInsertErrors = results.filter((result) => result.status === 'rejected');
+  if (dbInsertErrors.length > 0) {
+    // eslint-disable-next-line no-console
+    console.error(dbInsertErrors);
+  }
 
   dbMutex = false;
 
--- a/src/alerting.ts
+++ b/src/alerting.ts
@@ -13,7 +13,6 @@ import { BeaconSetTrigger, BeaconTrigger } from './validation';
 import { calculateUpdateInPercentage } from './calculations';
 import { UpdateStatus } from './check-condition';
 import { HUNDRED_PERCENT } from './constants';
-import { sleep } from './utils';
 import { getState } from './state';
 import { RateLimitedProvider } from './providers';
 import { logger } from './logging';
@@ -69,35 +68,14 @@ const writeRecords = async () => {
   ).filter((result) => result.status === 'rejected');
 
   if (results.length > 0) {
-    // eslint-disable-next-line no-console
-    console.error(results);
+    logger.error(`DB Writer error: ${results}`);
   }
 
   dbMutex = false;
 
   if (getState().stopSignalReceived) {
-    // eslint-disable-next-line no-console
-    console.log('Clearing DB writer interval due to stop signal received...');
+    logger.info('Clearing DB writer interval due to stop signal received...');
     clearInterval(dbWriterInterval);
-  }
-};
-
-let dbWriterInitialised = false;
-
-const dbWriterLoop = async () => {
-  if (dbWriterInitialised) {
-    return;
-  }
-
-  dbWriterInitialised = true;
-
-  dbWriterInterval = setInterval(writeRecords, 10_000);
-
-  if (process.env.DEBUG_DB_WRITER) {
-    setInterval(() => {
-      // eslint-disable-next-line no-console
-      console.log(`DB writer queue size: ${recordsToInsert.length} | ${(Date.now() - lastDbInsert) / 1_000}`);
-    }, 2_000);
   }
 };
 
@@ -136,11 +114,13 @@ export const getNodaryData = async (): Promise<NodaryData> => {
       axios({
         url: process.env.NODARY_DATA_URL,
         method: 'GET',
+        timeout: 10_000,
       }),
-    { retries: 3, attemptTimeoutMs: 14_900, totalTimeoutMs: 15_000 }
+    { retries: 0, attemptTimeoutMs: 14_900, totalTimeoutMs: 15_000 }
   );
+
   if (!nodaryResponse.success) {
-    await limitedSendToOpsGenieLowLevel(
+    limitedSendToOpsGenieLowLevel(
       {
         message: 'Error retrieving Nodary off-chain values in Airseeker Monitoring',
         priority: 'P4',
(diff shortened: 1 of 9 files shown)


src/alerting.ts
```diff
@@ -17,12 +18,83 @@ import { getState } from './state';
...
+const GATEWAYS_BAD_TRIES_AFTER_WHICH_CONSIDERED_DEAD = 3;
+
 export let prismaActive = prisma;
 
 export const setPrisma = (newPrisma: any) => {
   prismaActive = newPrisma;
 };
 
+type DbRecord = {
+  model: 'deviationValue' | 'rPCFailures' | 'compoundValues' | 'dataFeedApiValue' | 'gatewayFailures';
```
aquarat, line 30: I don't know if there's a better way of typing this or handling this.

src/alerting.ts
```diff
@@ -7,22 +7,78 @@ import axios from 'axios';
...
+
+export const clearRecordsToInsert = () => {
+  recordsToInsert = [];
+};
+
+export const getRecordsToInsert = () => recordsToInsert;
+
+const addRecord = (record: DbRecord) => {
+  // eslint-disable-next-line functional/immutable-data
+  recordsToInsert.push(record);
```
aquarat, line 46: We batch pending DB inserts now, which speeds things up and reduces DB load.

src/alerting.ts
```diff
@@ -7,22 +7,78 @@ import axios from 'axios';
...
+  lastDbInsert = Date.now();
+
+  const bufferedRecords = groupBy([...recordsToInsert], 'model');
+  recordsToInsert = [];
+
+  const results = (
+    await Promise.allSettled(
+      Object.entries(bufferedRecords).map(([model, data]) =>
+        // @ts-ignore
+        prismaActive[model].createMany({ data: data.map((item) => item.record) })
```
aquarat, line 65: Not sure if this can be done better...

src/alerting.ts
```diff
@@ -7,22 +7,78 @@ import axios from 'axios';
...
+      )
+    )
+  ).filter((result) => result.status === 'rejected');
+
+  if (results.length > 0) {
+    logger.error(`DB Writer error: ${results}`);
```
aquarat, line 71: If we drop a sample it's not a big deal and if we drop everything, well, bigger issue that will be obvious.

src/alerting.ts
```diff
@@ -58,11 +114,13 @@ export const getNodaryData = async (): Promise<NodaryData> => {
...
         method: 'GET',
+        timeout: 10_000,
       }),
-    { retries: 3, attemptTimeoutMs: 14_900, totalTimeoutMs: 15_000 }
+    { retries: 0, attemptTimeoutMs: 14_900, totalTimeoutMs: 15_000 }
   );
+
   if (!nodaryResponse.success) {
-    await limitedSendToOpsGenieLowLevel(
+    limitedSendToOpsGenieLowLevel(
```
aquarat, line 123: Sometimes we are limited by OG's rate limiting, in those case we don't want it to slow down the entire app. This was causing samples to be dropped.

src/alerting.ts
```diff
@@ -189,6 +231,19 @@ export const recordRpcProviderResponseSuccess = async (contract: Api3ServerV1, s
   }
 };
 
+export const getBaseUrl = (fullUrl: string) => {
```
aquarat, line 234: We use the base URL for (1) finding gateways and (2) hashing the gateway URL so it can be looked up later by Mertcan as part of his fault tracing.

src/alerting.ts
```diff
@@ -222,62 +277,70 @@ export const recordGatewayResponseSuccess = async (templateId: string, gatewayUr
...
+    .map((gateway) => findGateway(airnodeAddress, gateway.url)?.badTries ?? 0)
+    .filter((badTries) => badTries > GATEWAYS_BAD_TRIES_AFTER_WHICH_CONSIDERED_DEAD).length;
+
+  if (newGatewayResultStatus.badTries > GATEWAYS_BAD_TRIES_AFTER_WHICH_CONSIDERED_DEAD) {
+    limitedSendToOpsGenieLowLevel(
       {
         message: `Dead gateway for Airnode Address ${airnodeAddress}`,
         priority: 'P3',
-        alias: `dead-gateway-${airnodeAddress}${generateOpsGenieAlias(gatewayUrl)}`,
+        alias: `dead-gateway-${airnodeAddress}${generateOpsGenieAlias(baseUrl)}`,
```
aquarat, line 294: We're only interested in the base URL, not the URL+template ID (which is what we used before).

src/constants.ts
```diff
@@ -46,5 +46,5 @@ export const DIRECT_GATEWAY_MIN_TIME_DEFAULT_MS = 20;
 export const DIRECT_GATEWAY_MAX_CONCURRENCY_DEFAULT = 10;
 
 // TODO: load these 2 from env var instead
-export const DATAFEED_READ_BATCH_SIZE = 100;
-export const DATAFEED_UPDATE_BATCH_SIZE = 10;
+export const DATAFEED_READ_BATCH_SIZE = 1;
```
aquarat, line 49: Reading more than 1 has been unreliable on Kava - and I'm tired of troubleshooting this app - if you can see why it's failing, LMK.

src/fetch-beacon-data.ts
```diff
@@ -92,7 +78,7 @@ export const fetchBeaconData = async (beaconId: string) => {
   }
 
   const goRes = await go(fetchFn, {
-    retries: 2,
+    retries: 0,
```
aquarat, line 81: We don't need/want it to retry because if it fails we'll try the whole thing again very soon.

src/main.ts
```diff
@@ -16,6 +16,10 @@ export const handleStopSignal = (signal: string) => {
   expireLimiterJobs();
   updateState((state) => ({ ...state, stopSignalReceived: true }));
   heartbeatReporter(getState().config);
+
```
aquarat, line 19: Very annoying when the app doesn't quit quickly in response to ctrl+c. This fixes that.

src/alerting.ts
vponline: I like this solution 😄, anything else would be a much longer utility function

vponline: the createMany's are quite fast so this could even be made sequential if we find DB load to bee still too much later on 🤷

src/alerting.ts
vponline: I can't think of anything that would be more elegant than this either (reminds me of the old telemetry 😄)

src/constants.ts
vponline: wow! This would be a problem for the main airseeker as well, right?

vponline (approved):
LGTM, I didn't see any issues 🚀

PR #493, opened 2023-10-24
Stop treating `mantle` particularly

### LINKED ISSUE #492: Fix miscalculation of required gas price while filtering wallets for mantle
While filtering wallets for Mantle variants, rollup_gasPrices method was being used, which returns the following response:

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "l1GasPrice": "0xbd06339caf08",
        "l2GasPrice": "0x2faf080"
    }
}
```

Then summed both `l1GasPrice` and `l2GasPrice` and multiplied the result by `fulfillmentGasLimit` to calculate the required gas to update the beacon. However, it is crucial that the calculation be done as follows:

```
l1GasPrice * l1GasUsed + l2GasPrice * l2GasUsed
```

This is because `l1GasUsed` is much smaller than `l2GasUsed` (which is equivalent to the `fulfillmentGasLimit`). Failing to consider this difference can result in filtering sponsored wallets as not funded, even when they are funded. This can lead to catastrophic issues, particularly in terms of data feed updates.

My proposed solution is not treating mantle specifically and only depending on l2 gas price because there is no way to get accurate `l1GasUsed` for the transaction. This can cause filtering not funded wallet as funded but this is not an issue other than extra RPC calls.

bdrhn9:
Closes #492

### DIFF
--- a/src/wallets.test.ts
+++ b/src/wallets.test.ts
@@ -272,31 +272,6 @@ describe('hasEnoughBalance', () => {
     );
   });
 
-  it('should throw an error when failed to get gas price (mantle)', async () => {
-    const sponsorWallet = {
-      provider: {
-        getBlock: jest.fn().mockResolvedValue({
-          timestamp: Math.floor(Date.now() / 1000),
-        }),
-        network: {
-          chainId: 5000,
-          name: 'mantle',
-        },
-        send: jest.fn().mockImplementation((method: string) => {
-          if (method === 'rollup_gasPrices') {
-            throw new Error('getGasPrice: Unexpected');
-          }
-        }),
-        getGasPrice: jest.fn().mockResolvedValue(ethers.utils.parseUnits('10', 'gwei')),
-      },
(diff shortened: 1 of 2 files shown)


bdrhn9:
This is important fix for the future deployments. @aquarat please deploy Airseekers after this PR merged for incoming orders (because currently we have no managed dapi on `mantle` it's okay to not re-deploy existing ones)

acenolaza (approved):
LGTM 👍🏻 (should we maybe add an small extra % to try to account for the L1 gas usage?)

bdrhn9:
> LGTM 👍🏻 

Thank you for the prompt review, which was greatly appreciated due to its urgency.

> (should we maybe add an small extra % to try to account for the L1 gas usage?)

I think you mean extra % L1 gas usage for only mantle chains. However, I recommend avoiding complexity because it broke us yesterday 😄 . The only potential drawback of not including L1 gas costs is that it may misclassify non-funded wallets as funded and it won't create issue rather than extra failing call to update data feed.

PR #499, opened 2023-11-23
Increase limits on OpsGenie integration

aquarat:
Closes https://github.com/api3dao/tasks/issues/380

This PR increases the Bottleneck limits on OpsGenie API calls. The default is a concurrency of 1 with a minTime of 500ms, which is effectively 2 Hz. 

[A few weeks later(!)] This has been running for some time now and everything looks good, sao safe to merge.


aquarat:
This has been running for ages, so merging as I need to also test the other branch (that replaces nodary references).

PR #501, opened 2023-12-08
Update missing value alert

### LINKED ISSUE #500: [Airseeker Telemetry]: Swap out/remove error message Nodary reference
See https://api3workspace.slack.com/archives/C05RMQJ3X2L/p1701179579012749?thread_ts=1701177769.730009&cid=C05RMQJ3X2L

The Airseeker Telemetry variant creates OpsGenie alerts citing "Nodary" when it's referring to reference datafeed values. This is actually incorrect and confuses Nodary respondents (Bedirhan). It should say something else when referring to these reference values, possibly "API Analysis" or perhaps just remove the Nodary name entirely.

Probably name it "API-Analysis" or "reference data provider".

vponline:
Closes [[Airseeker Telemetry]: Swap out/remove error message Nodary reference#500](https://github.com/api3dao/airseeker/issues/500)

### DIFF
--- a/src/alerting.ts
+++ b/src/alerting.ts
@@ -122,16 +122,16 @@ export const getNodaryData = async (): Promise<NodaryData> => {
   if (!nodaryResponse.success) {
     limitedSendToOpsGenieLowLevel(
       {
-        message: 'Error retrieving Nodary off-chain values in Airseeker Monitoring',
+        message: 'Error retrieving off-chain values in Airseeker Monitoring',
         priority: 'P4',
-        alias: 'api3-nodaryloader-index-retrieval-airseeker-monitoring',
+        alias: 'api3-dataloader-index-retrieval-airseeker-monitoring',
         description: [`Error`, nodaryResponse.error.message, nodaryResponse.error.stack].join('\n'),
       },
       opsGenieConfig
     );
-    throw new Error('Error retrieving Nodary off-chain values endpoint');
+    throw new Error('Error retrieving off-chain values endpoint');
   }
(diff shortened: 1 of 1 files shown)


bdrhn9:
There are lots of place still Nodary naming populated here some examples:

- https://github.com/api3dao/airseeker/blob/8267a3c95e964d2763f9ab7adea800129dbc6d6d/src/alerting.ts#L125
- https://github.com/api3dao/airseeker/blob/8267a3c95e964d2763f9ab7adea800129dbc6d6d/src/alerting.ts#L350
- https://github.com/api3dao/airseeker/blob/8267a3c95e964d2763f9ab7adea800129dbc6d6d/src/alerting.ts#L377


Also in my opinion variables named as `*nodary*` that refers reference data should be changed. I don't think variable name changed is urgent but the alerts are.

vponline:
I see thanks, I will go through the rest of the alerts as well then 😄

aquarat (changes requested):
Looks fine 😆 but related to Bedirhan's comment it's probably safe to do a project-wide replace of nodary with maybe "source" 🤷 ?

vponline:
> Looks fine 😆 but related to Bedirhan's comment it's probably safe to do a project-wide replace of nodary with maybe "source" 🤷 ?

We could almost do this, there are a few db column names still containing nodary, but that should be ok.

aquarat:
I'll try and deploy this and see what happens, can always revert.

(update) Deployed... now we wait ~a day.

aquarat:
No errors and/or issues, so this is running nicely.

